PR #8, opened 2016-01-11
Make market transport water

asl97:
the ground work is already there (`MAX_WATER_IN_MARKET`), it just missing this for market to transport it.

note: i know the commit name sucks, i can change and rebase it if someone can suggest a better name.


uafr:
I'm not sure if this was intended. Are markets supposed to transport water?
There is no option in the market dialogue for water.

Sorry about the delay, looks like I had some notifications turned off.

asl97:
@uafr the fact that `MAX_WATER_IN_MARKET` exist must mean something.

PR #16, opened 2017-04-19
Fix integer overflow in tech_bonus calculation

MarcSchnyder:
In:
this->tech_bonus = (tech_level * ORGANIC_FARM_FOOD_OUTPUT) / MAX_TECH_LEVEL;
an overflow would occur. my tech level was ~ 390 (3'900'000 as stored in the int) while ORGANIC_FARM_FOOD_OUTPUT is 550.  which results in a number of 2'145'000'000 , which is unfortunately somewhat larger than an integer can represent. The number would therefore overflow and become something extreme negative. Some other code would then set the foodprod to 30 in organic_farm.cpp if it is lower than that , which explains why it was at least 30.
Casting tech_level to a long long int causes the calculation to be done with 64 bit integers and fixes the problem. After Dividing by MAX_TECH_LEVEL (1000000) the number is again safe to be stored as int.
funnily a farm without power would output ~135 which is the default and independent of the tech level , but with power output has gone down to 30 :D

### DIFF
--- a/src/lincity/modules/organic_farm.h
+++ b/src/lincity/modules/organic_farm.h
@@ -78,7 +78,7 @@ public:
         init_resources();
         this->tech = tech_level;
         setMemberSaved(&this->tech, "tech");
-        this->tech_bonus = (tech_level * ORGANIC_FARM_FOOD_OUTPUT) / MAX_TECH_LEVEL;
+        this->tech_bonus = int( ((long long int)tech_level * ORGANIC_FARM_FOOD_OUTPUT) / MAX_TECH_LEVEL );
         setMemberSaved(&this->tech_bonus, "tech_bonus");
         this->crop_rotation_key = (rand() % 4) + 1;
         this->month_stagger = rand() % 100;


MarcSchnyder:
According to: https://en.wikipedia.org/wiki/C_data_types , a "long int" is the same as an int , 32 bit , while only a "long long" , or "long long int" is 64bit. This may differ a bit per platform , according to wikipedia "long long int" seems to be the only suitable 64 bit int standard type.

So we cast from int (32 bit) to "long long int" (64 bit) so C uses large enough datatypes to calculate stuff, as it uses 64 bit operations if one of the inputs is 64 bit.

Other datatypes may be suitable as well. I dont know what would be the favourite choice of lincity devs :)
casting to float or double would work just as well. Im open for suggestions.


Btw i need to correct my initial description: a techlevel of 390 is not stored as 390'000  , but 3'900'000 , so factor 10000 not 1000 ,  as with 390'000 everything would work , which puzzled me for a minute :D

MarcSchnyder:
have now corrected my initial description in place

PR #17, opened 2017-04-28
Tell the compiler to use C++14

vorot93:
Clang does not build against C++14 by default. Therefore we should explicitly tell the compiler which version of C++ standard to use.

### DIFF
--- a/Jamrules
+++ b/Jamrules
@@ -11,6 +11,8 @@ if ! $(JAMCONFIG_READ)
   EXIT "Couldn't find config. Please run 'configure' first." ;
 }
 
+CXXFLAGS += -std=c++14 ;
+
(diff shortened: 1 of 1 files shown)


uafr:
Is this required to build with clang? I don't really like the idea of forcing some version because then we'll probably overlook new issues when C++17 hits distributions. But this is still better than conditional code for auto_pt/unique_ptr.

vorot93:
We should code and build against the current stable C++ if it is provided by LTS distributions. C++14 fits the bill today. C++17 will be viable when Ubuntu 18.04 LTS gets released.

And in any case there is no real reason for compatibility with older standards. Ubuntu Precise has reached EOL on 28 April and (afaik) Trusty's toolchain builds C++14 code just fine.

PR #22, opened 2017-08-16
Added Virtaal to the list of recommended translation editors.

gunchleoc:


vorot93:
I think we should do away with this kind of list completely and let the user decide which editor to use.

gunchleoc:
I still think it's a good idea to have it in for people who are new to translating, before that start hacking in Notepad or something, because they don't know that proper tools exist.

vassilis-papanikolaou:
This is a 'suggestion', ultimately the user is free to decide which editor to use.

dbear496:
I propose that we remove the list and instead provide a link to https://www.gnu.org/software/trans-coord/manual/web-trans/html_node/PO-Editors.html which is a list of PO editors recommended by the GNU project.

dbear496:
This has been superseded by PR #94.

PR #43, opened 2020-03-04
Added Haiku support

Crestwave:


dbear496:
Since PR #74, Lincity-NG uses `PHYSFS_getPrefDir` to get the configuration directory in a platform-independent manner. This may have fixed your issue with running on Haiku. But if you are still having issues with the location of the configuration directory on any platform, then it is likely an issue with the PhysFS library.

PR #50, opened 2021-04-26
Fix birth rate calculation

v1993:
Before, addition was done first, resulting in very wrong results.


v1993:
Should have merged this far earlier...

PR #51, opened 2021-05-12
Bug fixes for rocket pads and waste burning.

ever-3W:
Bugs fixed:
1. Rocket pads could accumulate far more than 100% waste, because there was no check to prevent this. A check for the waste level was added to prevent this - now rocket pads may not work if there is too much waste.
2. Fires from waste burning ("waste" fires) caused intermittent segfaults because the game tries to delete them twice.
3. Rocket launch dialog did not function correctly: selecting "No" if asked to launch caused the rocket to disappear without registering a succesful/failed launch.


v1993 (approved):
Looks good, thanks!

PR #53, opened 2021-06-21
ax_check_gl.m4: Also check against -lOpenGL

lanodan:
This allows lincity-ng to compile on wayland without libX11.

Note: This patch also works on lincity-ng 2.5-beta


dbear496:
The autoconf/jam build system was recently replaced with cmake, so this PR is no longer applicable.

PR #56, opened 2021-11-25
Fix compile error in release build

demotomohiro:
I got compile error when I tried to build lincity-ng in release build.

`src/lincity-ng/Debug.hpp` defines some macros and functions for debuging.
But they are not defined in release build because `NDEBUG` is defined and cause compile error.

macros and classes defined in Debug.hpp are not used and defining only `checked_cast` fixed the compile error.


v1993:
Nice, thanks a lot!

PR #66, opened 2022-12-05
Mouse gesture upgrades

dbear496:
A few of the mouse gestures have been really annoying, so this PR changes some of them:

1) Viewport panning is now done by dragging with middle click. Also the direction is changed so it feels more like dragging the map instead of guiding the viewport. I find it way more intuitive to use this way. Also, when dragging, the cursor changes to a hand (previously, the cursor disappeared while dragging the map). Now the mouse wheel can control all map navigation i.e. both panning and zooming.

2) Right click shows information about a building. I find this useful to avoid switching back and forth between the query tool so much. This replaces the previous middle click action which would show information about the tile underneath a building. Viewing info about the tile underneath a building can be toggled with the 'G' key.

3) When zooming with the mouse wheel, the zoom is relative to the pointer. That is, it will now zoom in towards the pointer or out away from the pointer. This is especially useful when zooming in on a specific area. Previously, panning was often required after zooming in.

4) Scrolling by moving the pointer to the screen edge is disabled when the mouse leaves the window. Previously, the map would start scrolling uncontrollably whenever I moved the pointer outside the window.

### DIFF
--- a/README
+++ b/README
@@ -157,6 +157,7 @@ use shift + direction to scroll faster
  h  to hide high buildings. Press h again to show them.
  v  to cycle through MiniMap-overlay modes
  b  toggle between current tool and bulldoze mode 
+ g  toggle between right click showing building or tile info
  F1 Help
  
  F12 quick save
--- a/README
+++ b/README
@@ -114,13 +114,12 @@ any:
  Click on Minimap shows the selected Area on mainscreen.
 
 right:
- Drag mainscreen with right mousebutton.
- right click on mainscreen centers to tile under cursor.
+ right click on mainscreen to show info on building under cursor.
  select tool from menu root
  show help on tool
 
 middle:
- middle click on mainscreen to show information about area under cursor.
+ pan map on mainscreen.
 
 left:
  Perform action depending on selected tool. Bulldoze, show Information,
@@ -175,4 +174,3 @@ You can contact us at the lincity-ng-devel mailinglist:
 or you might be able to catch us in irc at irc.freenode.net #lincity.
 
 Visit our homepage: https://github.com/lincity-ng/lincity-ng
-
--- a/src/gui/Event.cpp
+++ b/src/gui/Event.cpp
@@ -52,6 +52,18 @@ Event::Event(SDL_Event& event)
             type = MOUSEWHEEL;
             scrolly = event.wheel.y;
             break;
+        case SDL_WINDOWEVENT:
+            switch(event.window.event) {
+            case SDL_WINDOWEVENT_ENTER:
+                type = WINDOWENTER;
+                break;
+            case SDL_WINDOWEVENT_LEAVE:
+                type = WINDOWLEAVE;
+                break;
+            default:
+                type = WINDOWOTHER;
+            }
+            break;
         default:
             assert(false);
     }
@@ -63,4 +75,3 @@ Event::Event(float _elapsedTime)
 }
 
 /** @file gui/Event.cpp */
-
--- a/src/gui/Event.hpp
+++ b/src/gui/Event.hpp
@@ -48,6 +48,12 @@ public:
         MOUSEBUTTONUP,
         /// a mouse wheel has been turned
         MOUSEWHEEL,
+        /// window gained mouse focus
+        WINDOWENTER,
+        /// window lost mouse focus
+        WINDOWLEAVE,
+        /// other window event
+        WINDOWOTHER,
     };
     /// Create an update Event
     Event(float elapsedTime);
@@ -78,4 +84,3 @@ public:
 
 
 /** @file gui/Event.hpp */
-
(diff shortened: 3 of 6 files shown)


v1993:
Quite nice changes overall, although I think a little tweaking is needed:

> This replaces the previous middle click action which would show information about the tile underneath a building (which I found quite useless).

I remember this being very useful to check pollution levels under the building, so it would be nice to preserve this action.

> Scrolling by moving the pointer to the screen edge is disabled when not in full-screen mode. Previously, the map would start scrolling uncontrollably whenever I moved the pointer outside the window.

I think it would be better to stop scrolling when mouse pointer leaves the window. I find scrolling this way to be super convenient and current check does not account for mouse leaving game window while full-screen (which easily happens on systems with multiple monitors).

dbear496:
Thanks for the feedback; I think you brought up some good points.

About checking pollution levels underneath a building. I usually use the pollution overlay to get a good idea of where pollution is a problem. Or I will look at the unnatural death rate on a residence. For this reason, I thought the old middle-click feature was not useful -- or at least not useful enough to have a dedicated mouse button.

But in the interest of not removing an existing feature, I have an idea: a right click can show tile information while the button is held down, and then show building information once the button is released. Also, I can allow the information to change if the mouse moves to a new tile while the button is held down. (Previously, moving to a new tile while holding middle click would not change for which tile the info is shown.)

About scrolling by moving the mouse to the screen edge, I agree that this needs to be disabled when the mouse moves outside the window in fullscreen mode. However, I have mixed feelings about having it enabled in windowed mode. For one, in windowed mode it could be difficult to use; that is, it could be difficult to put the pointer close to the edge of the window without going outside the window. Well, 'difficult' may be a strong word, but I does require being intentional about where the pointer is, and it isn't like in fullscreen where you can just smash the cursor against the edge of the screen. Also, I have many times scrolled accidentally while trying to interact with buttons on the edge of the screen. You said it is very handy to scroll by moving the pointer to the edge (and I agree for fullscreen), but do you use this feature in windowed mode? If you think it is still useful in windowed mode, then I will enable it in windowed mode and just disable it based on whether the pointer is inside the window. (But I will still disable it for my private build, so I don't keep scrolling accidentally.)

dbear496:
I tried to use `event.inside` to determine whether the mouse is inside the window for a MOUSEMOTION event, but it seems that no event is produced while the mouse is outside the window. This prevents the `event.inside` flag from being activated, so it is not very useful. I noticed one exception, and that is at the bottom of the window the flag is activated, and this prevents scrolling down. I noticed the same behavior in fullscreen with two monitors -- moving to the bottom of the screen did not scroll because the pointer went outside, and moving up to my second monitor did not stop scrolling.

dbear496:
Did you know the 'g' key already shows information about the tile underneath a construction like the old middle click? Do you still think a mouse button should do this?

I tried making the right mouse button behave as I described earlier: hold shows info on tile, release shows info on building. But this makes the MPS window flash when the button is clicked.

v1993:
More about scrolling - yeah, I use it in windowed mode too. It seems that you can use `SDL_WINDOWEVENT_ENTER`/`SDL_WINDOWEVENT_LEAVE` events to track mouse state. Maybe it would make sense to add a setting to toggle it on/off? Might be better done as a separate follow-up PR tho.

And on tile info with right-click - how hard would it be to make repeated right clicks page trough building/tile info pages? I feel like that would feel the most natural.

dbear496:
Thanks, I'll look into the `SDL_WINDOWEVENT_ENTER`/`SDL_WINDOWEVENT_LEAVE` events. It might be worth updating all uses of `event.inside` because it isn't working properly.

Good idea about paging through building/tile info. However, I obviously think building info should be the first page, and you probably think tile info should be the first page.

dbear496:
This will require a restructuring of the current MPS paging system. As it stands, the number of pages is 2 for all types of buildings, and the second page is only used by track, road, and rail -- other pages just put the same content on both pages. So when I try adding a page for tile info, you have to click three times to get to it even if the building only needs one page. And on top of that, it only shows tile info for the top corner of the building.

None of this is not fixable, but it will require a bit more effort.

An alternative solution may be to page with the g key. This can be set up so that you can press the g key once and then you will always get tile info. Then it won't be such a big deal which page is first.

v1993:
I'll hopefully take another look later, but for now - I'm a little concerned about your usage of SDL's cursor API. It probably leaks memory for now (there are no `SDL_FreeCursor` calls). How about caching all used cursors and freeing them on shutdown instead of creating a new cursor object every time?

dbear496:
Oh, thank for pointing out that memory leak. That was careless on my part. I believe I have fixed it now. I wasn't exactly sure where to put the interned cursor. It could realistically go in main, Game, GameView, or Window. I put it in GameView for now.

dbear496:
@uafr do you have any comments about this this PR? Or PRs #64 and #68?

dbear496:
There are now two PRs (#68 and #72) that depend on this PR, so it'd be nice to get this merged.

dbear496:
Hey, @v1993, any chance this could take another look at this?

PR #68, opened 2022-12-14
Scrolling improvements

dbear496:
This implements various improvements to the scrolling behavior and hitting the map edge.

- **Adds scrolling with WASD.**
- **Tweaks the scrolling speed in different directions when using the keyboard.** Now, scrolling vertically, horizontally, and diagonally feels more like the same speed. Previously, horizontal scrolling felt faster than vertical scrolling due to the distortion of the tiles, and diagonal scrolling was substantially faster because it added horizontal and vertical scrolling with no normalization.
- **Tweaks the scrolling speed with the keyboard at different zoom levels.** Now, scrolling feels appropriately faster when zoomed out.
- **Anchors the cursor when panning, and keeps the anchor position when hitting the map edge.** Previously, the anchor position would change when pushing against the map edge.
- **Removes lag when scrolling against the map edge.** Previously, hitting the edge would cause the draw function to skip a beat (or a lot more beats).
- **Makes scrolling against the map edge smoother.** Previously, the logic which reset the viewport position would move the viewport to the center of an edge tile, and this made the viewport jump a half tile. Now, the viewport center is constrained to and reset to the edge of the outside tile to avoid the jump.
- **Removes the teleport effect of zooming in and out.** Previously, when the map edge was hit while zooming out, it would shift the viewport and consequently the cursor so that zooming back in would not go to the same place. A similar effect could happen when zooming in and then out. This was particularly annoying when the zoom was accidental because zooming the opposite direction would not get back to the same position. Now, if the viewport is shifted from hitting the map edge, then zooming in the opposite direction will un-shift it. Moving the mouse will cancel the saved shift.
- **Persists the shift key when hitting the map edge.** Previously, when the map edge was hit with the shift key pressed, the game would forget that the shift key was pressed -- shift would have to be released and pressed again to re-enable fast scrolling. Now, this is fixed.

**Note: This PR is based on commits from PR #66, so this should not be merged until after PR #66 is merged.** Otherwise, this is ready for review even though it is marked as a draft.

### DIFF
--- a/README
+++ b/README
@@ -114,13 +114,12 @@ any:
  Click on Minimap shows the selected Area on mainscreen.
 
 right:
- Drag mainscreen with right mousebutton.
- right click on mainscreen centers to tile under cursor.
+ right click on mainscreen to show info on building under cursor.
  select tool from menu root
  show help on tool
 
 middle:
- middle click on mainscreen to show information about area under cursor.
+ pan map on mainscreen.
 
 left:
  Perform action depending on selected tool. Bulldoze, show Information,
@@ -158,6 +157,7 @@ use shift + direction to scroll faster
  h  to hide high buildings. Press h again to show them.
  v  to cycle through MiniMap-overlay modes
  b  toggle between current tool and bulldoze mode 
+ g  toggle between right click showing building or tile info
  F1 Help
  
  F12 quick save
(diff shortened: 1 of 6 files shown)


dbear496:
I'm marking this ready for review (because it really is) even though it depends on #66 which is _still_ not merged.

PR #72, opened 2022-12-26
Fix scroll wheel focus

### LINKED ISSUE #71: Scroll wheel performs multiple actions simultaneously
When the minimap is in the transportation tab, then the scroll wheel will both zoom the mainscreen and cycle through commodities in the minimap no matter where the mouse is. So even when the mouse is on the minimap, the scroll wheel will zoom the mainscreen. And even when the mouse is on the mainscreen, the scroll wheel cycles through commodities in the minimap.

Similar behavior happens when a scrollable dialog is open where the scroll wheel simultaneously scrolls the dialog and zooms the mainscreen no matter where the cursor is.

This behavior is very annoying. The scroll wheel should only perform one action at a time, and that should be based on the mouse position.

dbear496:
This fixes #71.

- On scroll wheel events, components now get an accurate indication of whether the mouse is inside the component.
- Scroll wheel action now only applies to the component that the mouse is over.
- This uses a feature from SDL 2.26.0 to get the mouse position for scroll wheel events. For users with an older version, this is detected in ./configure, and a different method is used to get the mouse position.

**Note: This PR is based on commits from PR #66. So this should not be merged until after PR #66 is merged.** Otherwise, this is ready for review even though it is marked as a draft.

### DIFF
--- a/README
+++ b/README
@@ -157,6 +157,7 @@ use shift + direction to scroll faster
  h  to hide high buildings. Press h again to show them.
  v  to cycle through MiniMap-overlay modes
  b  toggle between current tool and bulldoze mode 
+ g  toggle between right click showing building or tile info
  F1 Help
  
  F12 quick save
--- a/README
+++ b/README
@@ -114,13 +114,12 @@ any:
  Click on Minimap shows the selected Area on mainscreen.
 
 right:
- Drag mainscreen with right mousebutton.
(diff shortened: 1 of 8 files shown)


dbear496:
I'm marking this ready for review (because it really is) even though it depends on #66 which is _still_ not merged.

Issue #73, opened 2022-12-26
'jam -a' overwrites changes to data/gui/tradedialog.xml


dbear496:
It appears that 86694fc1 adds some waste buttons to the market trade dialog. However, this was done by editing `data/gui/tradedialog.xml` which prominently says at the top of the file "**WARNING: This file is automatically generated. Do not edit!**". If I run `jam -a`, then the changes are overwritten, and git status reports `data/gui/tradedialog.xml` modified. Instead, I believe the waste buttons should be added in `data/gui/dialogs/tradedialog.xml`.

But this leads to another issue: Why are automatically generated files in the git tree at all? I think the generated files should be added to gitignore.

dbear496:
This is fixed in PR #75.

PR #74, opened 2022-12-28
Update Physfs deprecations

dbear496:
This attempts to remove uses of deprecated functions from the PHYSFS library. This is based on PR #42 with some improvements:

- Resolves merge conflicts with master.
- Fixes indentation issues from #42.
- Improves the error message when a file fails to stat.
- Undoes changes to 'data/gui/tradedialog.xml' that were caused by issue #73.
- Makes some error-checking code more concise.
- Fixes a few missed `!` (logical negation) from #42.
- Updates readme to reflect the new PHYSFS version requirement.

I would contribute to #42 instead of opening a new PR, except the author of #42 seems to have deleted their GH account and that branch is not writable by me.


dbear496:
I am not sure why `initPhysfs` in src/lincity-ng/main.cpp mounts everything manually. It seems that it could simply call `PHYSFS_setSaneConfig` instead. Why is this?

dbear496:
The changes to src/lincity-ng/main.cpp introduced by 63aa59c857972f8681115d9eb3b2058c05af3190 make absolutely no sense. There is a distinction between `defined(__APPLE__)` and `__APPLE__ == 0` that doesn't look right at all.

dbear496:
It appears that `initPhysfs` used to use `PHYSFS_setSaneConfig`, and 240371169207d786483bed7f1de22b3ce984e3c7 changed it to exclude the baseDir. I am not sure why the base dir needs to be excluded, and there is no indication of why the change was made either in comments in the code or in the commit message.

dbear496:
Now, running `jam -a` should result in only one PHYSFS deprecation warning, and this is for `getUserDir`. This is used to get the path names for the old configuration directories which are in the user's home directory. The new configuration directory is provided by PHYSFS_getPrefDir per updated PHYSFS recommendation.

v1993:
IMO it would be a good idea to make a simple wrapper function for `PHYSFS_getErrorByCode(PHYSFS_getLastErrorCode())`.

dbear496:
I think you have a valid point. I have made that change in c9c9d361e1a6509e568050150531944c71f27399 which is on the master branch of my fork.

PR #82, opened 2023-11-24
Performance improvements

dbear496:
Provides a major performance improvement to `Construction::trade`, speeding up fast simulation mode by about 2x overall.

- Stores commodity counts and commodity rules in std::array instead of std::map to speed up access to these values.
- Eliminates some division operations.
- I measured a ~2x speedup on a Intel i7 Core processor. Now, one simulated year takes ~5 seconds.
- Does not change any functionality, i.e. trading still operates exactly as before.

With this massive speedup, animation (particularly power lines) is also faster. I think some future work should go toward normalizing animation speed and decoupling it from simulation speed.

**Note:** Depends on #42, #64, #66, #68, #72, #74, #75, #78, #80, #81.


dbear496:
For the benchmark I ran before, I was simulating a relatively large city. When simulating a very small city, I went through 203 simulated years in one minute.

PR #88, opened 2023-12-15
configure.ac: Fix incompatible pointer type in zlib probe

fweimer-rh:
The gzopen function returns type `gzFile`, and not `gzFile *` as previously assumed.  This error causes the probe to fail unconditionally even if zlib is available if the compiler treats such incompatible types as errors (because they are not valid C).

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -86,7 +86,7 @@ AM_BINRELOC
 
 NP_FINDLIB([ZLIB], [zlib], [zlib],
 	NP_LANG_PROGRAM([[#include <zlib.h>]],
-            [[gzFile* file = gzopen("", "");]]),
+            [[gzFile file = gzopen("", "");]]),
 	[], [-lz],
 	[],
 	[AC_MSG_ERROR([Please install zlib])])


dbear496:
I agree with you that this was an issue, and I appreciate the PR. However, I recently upgraded the project to use cmake instead of autoconf, so the issue doesn't exist anymore.

dbear496:
I will evaluate whether I can merge your changes into a different branch.

dbear496:
The cmake version uses `find_package` to detect the zlib library, so I do not believe it suffers the same issue.

dbear496:
Here is the relevant part of the zlib documentation.

> ```
> typedef struct gzFile_s *gzFile;    /* semi-opaque gzip file descriptor */
> ```
> ```
> ZEXTERN gzFile ZEXPORT gzopen(const char *path, const char *mode);
> ```

https://www.zlib.net/manual.html

dbear496 (approved):
Indeed `gzopen` returns type `gzFile` according to [the zlib documentation](https://www.zlib.net/manual.html).

PR #105, opened 2024-01-31
Git hash version

dbear496:
Uses `git describe` to generate the version number in order to avoid confusion about what version someone is using when they report an issue.


dbear496:
Many thanks to rpavlik for creating the [cmake-modules](https://github.com/rpavlik/cmake-modules) repository, which made incorporating the git hash into CMake very easy.

PR #127, opened 2024-04-12
Fix GUI XML install

### LINKED ISSUE #122: Couldn't display dialog 'tradedialog.xml': Couldn't find a component factory for 'Dialog'
version 2.10.1

1. starts Rocket98 map
2. double click on a market

I get:
`Couldn't display dialog 'tradedialog.xml': Couldn't find a component factory for 'Dialog'`

### LINKED ISSUE #125: Problem with ok_dial_box: Couldn't find a component factory for 'Invention'
version 2.10.1

- starts a new game, builds a monument, then use maximum speed
- a popup appears with an error like `Problem with ok_dial_box: 'cricketup.mes' + ""`

Also, on stdout:
`Problem with ok_dial_box: Couldn't find a component factory for 'Invention'`

dbear496:
Fixes the CMake `install` directives for GUI XML so that the xslt-processed XML is installed. Previously, the source XML was installed.

fixes #122, fixes #125

### DIFF
--- a/data/gui/CMakeLists.txt
+++ b/data/gui/CMakeLists.txt
@@ -24,15 +24,15 @@ set(guiSourceFiles
 
 foreach(guiSourceFile ${guiSourceFiles})
   add_custom_command(
-    OUTPUT ${guiSourceFile}
+    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${guiSourceFile}
     COMMAND LibXslt::xsltproc -o ${CMAKE_CURRENT_BINARY_DIR}/${guiSourceFile} ${guiSourceFile}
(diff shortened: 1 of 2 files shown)


FabioPedretti:
Nice, thanks.
Do you think the others CMakeLists.txt are OK?
Also note that the 2.10.1 version currently in Debian and Ubuntu use CMake ( https://tracker.debian.org/news/1516197/accepted-lincity-ng-2101-1-source-into-unstable/ ), and are then affected, maybe a 2.10.2 release could be released soon?

dbear496:
> maybe a 2.10.2 release could be released soon?

Thanks for reminding me; this fix definitely warrants a patch release. I have done so in PR #128.

PR #152, opened 2024-07-08
Fix blacksmith monthly update

### LINKED ISSUE #150: Sometimes percentages are over 100%
Version: 2.11.0

I noticed that sometime production percentages get over 100% (I even see over 400%). Also all the percentages of the building are the same (see the screenshot). If I save the game and then reload it, the percentages are <= 100% and they are different again, so it may be a temporary or a visualization issue.

over100

I am anyway attaching a savegame where the percentage of blacksmith 10 was 118%.
[2_Y00143_Tech011_Cash-005M_P02337.gz](https://github.com/user-attachments/files/16119452/2_Y00143_Tech011_Cash-005M_P02337.gz)

dbear496:
This fixes an issue where blacksmiths could skip the monthly update. When this happened, the busy counter and production counters would not be reset, so the construction could display more than 100% busy/production.

- fixes #150

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,18 @@
+## Unreleased
+
+###### Internal
+- Fixed a compile-time error in src/gui/event.cpp when building with
+  SDL <= 2.26.0.
+- Fixed a compile-time error in src/lincity/xmlloadsave.cpp when building with
+  older g++ version.
+- Fixed parsing boolean from XML save file.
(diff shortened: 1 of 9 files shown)


dbear496:
@uafr Bruh, why are you rebasing? The commit re-ordering caused a change to be lost, and now the history has duplicated commits. I'm going to force push to fix this history.

PR #164, opened 2024-07-22
Refactor Load/Save

### LINKED ISSUE #153: Rocket starting issue
Version: 2.11.0

After I launch a rocket, it looks like I cannot launch a new rocket (there is no more progress on the rocket).
However, if I save the game and then immediately reload it, the rocket is immediately ready to be launched.
Attaching the savegame reproducing  this.

[1_Y00296_Tech137_Cash+288K_P09481.gz](https://github.com/user-attachments/files/16202771/1_Y00296_Tech137_Cash%2B288K_P09481.gz)

### LINKED ISSUE #172: set proper camera position when starting a game
1. When using Continue, New Game or Load after starting Lincity, the camera is set at north west, over a black area, out of the map.
2. When returning to main menu after playing a game and using New Game or Load the camera is set in last position in previous game.
3. When returning to main menu after playing a game and using Continue the camera is set in last position in previous game.

Point 1 and 2 are not nice, they should both point at the center of the map.
3 is OK, though.

### LINKED ISSUE #205: Heap overflow in XMLloadsave::readPbar
[XMLloadsave::readPbar](https://github.com/lincity-ng/lincity-ng/blob/cc6f912128e5f389b8f5d89968a380258ebc6619/src/lincity/xmlloadsave.cpp#L1244) reads the ID of the pbar from XML and, without verifying that it is in-bounds, uses it to index into the `pbars` array. A malicious save file can use this to arbitrarily write to any address in memory.

dbear496:
Completely rewrites the load/save logic to be more robust and future-proof. Consequently, this updates the load/save format so the game will not load save games from the previous version. However, I created [two scripts](https://github.com/dbear496/lincity-ng/tree/0ff04a6508d2f1f353d0d68cae4b3dcb2836f35e/contrib/ldsv-format-convert) that will upgrade the format from the format used in version 2.11.x with binary saving and seed saving both disabled.

This PR adds a new dependency on libXML++ for reading the save games in XML. And libXML2 is used for writing the same games, but this dependency already existed for the GUI. The previous version used custom XML parsing logic which was error-prone, had limited extensibility, and had security vulnerabilities (#185, #205). Thus using a library for XML parsing is much more robust.

Consequently, this PR pushes the C++ standard to C++17 to support libXML++.

Due to this PR changing the load/save format in a non-compatible way, this is targeting merge into version 2.13.0. Version 2.12.0 is due to deprecate binary saving and seed saving, so all games saved with 2.12.x will be compatible with the upgrade scripts.

Fixes #153
Fixes #172
Fixes #205


FabioPedretti:
Is compatibility really important (other than the included scenarios)?
Maybe just target for 2.12 with a note that old savegames are no longer compatible?

dbear496:
Yes, compatibility is very important. Imagine going to build your city that you've been working on for several days (or weeks), and finding that it won't load because of a routine system update.

IMO the load/save logic needs a facelift anyway. It is based on a custom XML parser that is very sensitive to the whitespace/formatting/ordering of the XML. And the construction members saving is very sensitive to the number and order of saved members. In other words, it's fragile. I think ideally, this could be upgraded to use JSON format, which is more modern, and use an external library that actually adheres to a standard.

dbear496:
I made some major updates to this PR and edited the OP to reflect.

dbear496:
_sigh._ This is why I don't use Ubuntu. I agree the proper solution is to add libxml++5.0 for these distros. An alternative is to statically link libxml++5.0.

dbear496:
Yes, adding an option to statically link it would make libxml++ 5.0 only a build dependency instead of also a runtime dependency.

dbear496:
@FabioPedretti I updated the Dockerfile that I use to create Ubuntu-compatable builds so that it links libxml++ statically. You can take a look at it [here](contrib/build/Dockerfile) as a reference point.

I also added a `#ifdef` on the call to `xmlTextWriterClose` because evidently that function isn't available in earlier versions of libxml2 (and Ubuntu is stuck in the past).

a-detiste:
I've limited time for this... But Ubuntu maintainer are welcome to work directly on Debian packages.

a-detiste:
The real XZ guy that sent us Merge Resquest at Games Team was much more laconic

dbear496:
> OK, so I suppose the Debian package (and likely Fedora) should be modified and include and build both libxml++5 and lincity-ng (and remember to update libxml++5 when there are for example security fixes). At least until libxml++5 is available on Debian. Or is there another way?

I understand this adds significant work for maintainers to pull in and build libxml++ 5.0. Really the best solution is to add libxml++ 5.0 packages to distribution repositories, but aside from that, IDK. I avoid git submodules because it is so difficult to work with. And you already nixed CMake FetchContent because it downloads during the configure step. 

Maybe I'm being cold, but I don't think it's my problem. I'm willing to help out in ways I can, but I'm not willing to bloat the lincity-ng git repo with an entire library just because Debian doesn't package such library in their repository. That would implicitly make me responsible for the library where it should be just between the distribution maintainers and the library developers. And I'd rather the lincity-ng git repo contain lincity-ng and only lincity-ng.

I believe this is a temporary problem. Eventually libxml++2 will go obsolete, and eventually libxml++5 will be packaged in distribution repositories. libxml++ is after all the recommended way to use libxml2 with C++ according to the libxml2 website, so it's not like it's some niche library.

dbear496:
@FabioPedretti do you want to convert the starting scenarios to the new format so that you can GZ compress them with your utility that evidently gets better compression ratios?

PR #170, opened 2024-08-02
Fix various issues in help pages

FabioPedretti:


data/help/es/industyh.xml
dbear496, line 1: Why does this file need to be deleted?

data/help/gl/industyh.xml
dbear496, line 1: Again, why does industryh.xml need to be deleted?

dbear496 (commented):
Looks good. I just have that one question about the industryh.xml deletion.

data/help/es/industyh.xml
dbear496: Oh, I see now after reading your commit message.

dbear496 (approved):

PR #174, opened 2024-08-17
script used for the fixes in #170

FabioPedretti:


data/help_templates/find-src-in-translated-help-missing-in-reference.sh
```diff
@@ -0,0 +1,51 @@
...
+	exit
+fi
+cd $DATA_DIR
+
+for H in help help_templates
+do
+	echo
+	echo === directory: $H ===
+	cd $H
+	CONTENT_FILES=`find -type f -iname "*.xml" | xargs grep src | cut -d'"' -f 2 | sort | uniq`
```
dbear496, line 21: This doesn't look right for help_templates, since it doesn't contain any XML files.

data/help_templates/find-src-in-translated-help-missing-in-reference.sh
```diff
@@ -0,0 +1,51 @@
...
+	elif [ $H = "help_templates" ]; then
+		REF="orig"
+	else
+		echo error
+		exit
+	fi
+	echo reference directory: $REF
+
+	cd $REF
+	XML_FILES=`find -type f -iname "*.xml" | sort`
```
dbear496, line 34: Again, this isn't right for help_templates.

data/help_templates/find-src-in-translated-help-missing-in-reference.sh
```diff
@@ -0,0 +1,51 @@
...
+
+	for F in $CONTENT_FILES
+	do
+		_F=`echo $F | rev | cut -d'/' -f 1 | rev`
+		grep [/\"]$_F $XML_FILES > /dev/null
```
dbear496, line 39: Use `grep -m 1` for performance. Also consider adding `--` to explicitly separate the files to search from the regex.

data/help_templates/find-translated-help-page-differences.sh
```diff
@@ -0,0 +1,50 @@
...
+cd $DATA_DIR
+
+for H in help help_templates
+do
+	echo
+	echo === directory: $H ===
+	cd $H
+	LANGUAGES_DIRS=`ls -d */ | sed 's/\/$//'`
+
+	_ENG_TMP=`mktemp`
```
dbear496, line 25: Be sure to clean up temporary files afterward.

data/help_templates/find-translated-help-page-differences.sh
```diff
@@ -0,0 +1,50 @@
...
+	elif [ $H = "help_templates" ]; then
+		REF="orig"
+	else
+		echo error
+		exit
+	fi
+	echo "reference directory: $REF (`ls $REF/*ml | wc -l` files)"
+	ls $REF > $_ENG_TMP
+
+	_TMP=`mktemp`
```
dbear496, line 37: Be sure to clean up temporary files when they are no longer needed.

data/help_templates/find-translated-help-page-differences.sh
```diff
@@ -0,0 +1,50 @@
...
+
+	for F in $LANGUAGES_DIRS
+	do
+		if [ $F = $REF ] ; then continue ; fi
+		ls $F > $_TMP
+		echo
+		echo "$F (`ls $F/*ml | wc -l` files):"
```
dbear496, line 44: Don't print this header if no issues are found for this translation.

dbear496 (changes requested):
- I think these scripts should be combined into one that is named `find-help-translation-issues.sh`. At the very least, the files names should be made more concise.
- These scripts are very difficult to read, so please add comments.

data/help_templates/find-help-translation-issues.sh
```diff
@@ -0,0 +1,101 @@
...
+cd help
+# get a list of all the "src" filenames (usually png) inside all languages xml help files
+# this include the filename path
+CONTENT_FILES=`find -type f -iname "*.xml" | xargs grep src | cut -d'"' -f 2 | sort | uniq`
+
+cd en
+# get a list of xml files inside the reference (en) directory
+XML_FILES=`find -type f -iname "*.xml" | sort`
+
+for F in $CONTENT_FILES
```
dbear496, line 81: I think this for loop could be replaced by a 'comm' command.

dbear496 (approved):
This looks good.

data/help_templates/find-help-translation-issues.sh
dbear496: I think all of check 2 can be replaced with the following (untested)
```
comm -23 \
  <(find . -type f -iname "*.xml" | xargs grep src | cut -d'"' -f 2 | sort | uniq) \
  <(find en -type f -iname "*.xml" | xargs grep src | cut -d'"' -f 2 | sort | uniq)
```

dbear496 (approved):

PR #188, opened 2024-09-30
Credits

dbear496:
- Updates the credits screen with new contributors.
- Adds years to the credits screen.
- Adds a license notice to the credits screen.
- Updates license/copyright headers in source files.


uafr:
It a licence change from GPL2 to GPL2+ possible?

dbear496:
> It a licence change from GPL2 to GPL2+ possible?

I though it is already GPL2+.

dbear496:
> What about using SPDX license headers? https://spdx.dev/learn/handling-license-info/

I'm not totally sold on SPDX headers; I just don't see much benefit to them. Much of the code base already uses the full GPL2+ license header, so I prefer just sticking with that for now.

PR #207, opened 2024-10-29
fixup after #173

dbear496:
- restricts installed locales to *.po
- moves installation of info files to DOCDIR


dbear496:
I removed "draft" from the PR title. I put "draft" in the commit message to indicate that it is untested and hence may not even build. And I like to avoid rewriting history that is already public -- It is annoying for everyone else who may have based work on it, and I have way too much experience with people `git pull`ing the changes and merging the old history.

dbear496:
Okay, I get your meaning. For this PR I guess it shouldn't be an issue since probably no one else is working on it.

PR #208, opened 2024-10-29
cleanup some duplicates files

FabioPedretti:
Note: now there are some holes in existing sound files, e.g. there is Tip1 and Tip3, but no more Tip2, I think it is still better than renaming existing files.


dbear496:
This is when I wish we could use symlinks, but unfortunately Windows coughs at those.

dbear496 (approved):

PR #222, opened 2024-11-13
set CPack generator based on platform

dbear496:
- ZIP generator for Windows
- TGZ generator for all other platforms


dbear496:
There is some benefit to having the package compression the same as the load/save compression to reduce the dependency set. Should lincity-ng use XZ for load/save compression?

dbear496:
Okay, I guess it makes sense to use TXZ then. Looks like it gets ~10% smaller archive.

PR #223, opened 2024-11-14
CI: Add initial workflow

### LINKED ISSUE #103: use reproducible testing environment
This project needs a reproducible testing environment so that testing is not dependent on the configuration of the developer's system. Such a testing environment would have caught issues #99 and #100 before they ever got released to master.

I believe a reproducible testing environment could be created with Vagrant or Docker.

andy5995:
This simply adds [a workflow](https://docs.github.com/en/actions/use-cases-and-examples/creating-an-example-workflow) that will build and install lincity when a pull request is open or changes are pushed to master, and report an error if either one fails.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -1,0 +1,71 @@
+name: C/C++ CI
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches: master
+  pull_request:
+    branches: master
+
+jobs:
+  build:
+    name: ${{ matrix.name }}-${{ matrix.os }}-${{ matrix.cc }}
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [default]
+        os: [ubuntu-22.04, ubuntu-24.04]
+        cc: [gcc, clang]
+    runs-on: ${{ matrix.os }}
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set CC
+      run: |
+        CC=${{ matrix.cc }}
+        if [ "$CC" = "gcc" ] || [ -z "$CC" ]; then
+          CC=gcc
+          CXX=g++
+        else
+          CC=clang
+          CXX=clang++
+        fi
+        echo "CC=$CC" >> $GITHUB_ENV
+        echo "CXX=$CXX" >> $GITHUB_ENV
+
+    - name: Show info
+      run: |
+        export -p
+        echo '${{ toJSON(matrix) }}'
+
+    - name: Install dependencies
+      run: |
+        sudo apt update
+        sudo apt remove -y firefox
+        sudo apt upgrade
+        sudo apt install -y \
+          cmake \
+          gettext \
+          libphysfs-dev \
+          libsdl2-dev \
+          libsdl2-gfx-dev \
+          libsdl2-image-dev \
+          libsdl2-mixer-dev \
+          libsdl2-ttf-dev \
+          libxml++2.6-dev \
+          libxslt1-dev \
+          libxslt1.1 \
+          xsltproc \
+          zlib1g-dev
+
+    - name: Configure
+      run: |
+        cmake --version
+        cmake -B build
+
+    - name: Build
+      run: cmake --build build --parallel
+
+    - name: Install
+      run: sudo cmake --install build


andy5995:
The results can be viewed at https://github.com/andy5995/lincity-ng/actions

andy5995:
Fails on Ubuntu 20.04
[logs_30865517602.zip](https://github.com/user-attachments/files/17740746/logs_30865517602.zip)

Errors like:

```
-- Configuring done (1.7s)
CMake Error at mk/cmake/modules/SDL2/FindSDL2_gfx.cmake:219 (set_target_properties):
  The link interface of target "SDL2::SDL2_gfx" contains:

    SDL2::SDL2-static

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  CMakeLists.txt:31 (find_package)
```

Log attached.

Ubuntu 20.04 is supported by Ubuntu until April of 2025 so I assume you want to support it as well. But please do let me know if that's incorrect.

dbear496:
I _should_ build on Ubuntu 20.04. Did the SDL CMake library target change between 20.04 and 22.04?

andy5995:
I don't know.

I'm trying a bit of testing locally in a 20.04 docker container:

```
root@33bbee2faa06:/lincity-ng# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.6 LTS"
root@33bbee2faa06:/lincity-ng# cmake -B _container
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.21 or higher is required.  You are running version 3.16.3


-- Configuring incomplete, errors occurred!
```

The [GitHub runner is using cmake to 3.31.0](https://github.com/actions/runner-images/blob/67857a41a163b96452c4b5a16058935458d7679d/images/ubuntu/Ubuntu2004-Readme.md?plain=1#L78). During `apt upgrade` I see it's installing 3.16.3-1ubuntu1.20.04.1.  Maybe cmake 3.13.0 gets installed with snap when the runner image is created. https://github.com/actions/runner-images

I tried with snap in my docker container, but 

```
snap install cmake --classic
error: cannot communicate with server: Post "http://localhost/v2/snaps/cmake": dial unix /run/snapd.socket: connect: no such file or directory

# systemctl start snapd
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

dbear496:
I do not think this PR closes issue #103 because it only builds and installs the project but never runs the game. There is already a dockerfile in the contrib directory that will build the project for Ubuntu, but issue #103 asks for a _testing_ environment, not just a _build_ environment.

andy5995:
> I do not think this PR closes issue #103 because it only builds and installs the project but never runs the game.

I've updated the PR to correct my oversight. I'll work on that after this is merged.

>There is already a dockerfile in the contrib directory that will build the project for Ubuntu, but issue #103 asks for a _testing_ environment, not just a _build_ environment.

A workflow like this will test the build anytime a pull request is opened or master is pushed to. There are also [methods that will prevent it from triggering](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-filters) when changes are made only to documentation or data, etc. https://docs.github.com/en/actions/use-cases-and-examples/creating-an-example-workflow

andy5995:
> > I do not think this PR closes issue #103 because it only builds and installs the project but never runs the game.
> 
> I've updated the PR to correct my oversight. I'll work on that after this is merged.
> 
> > There is already a dockerfile in the contrib directory that will build the project for Ubuntu, but issue #103 asks for a _testing_ environment, not just a _build_ environment.
> 
> A workflow like this will test the build anytime a pull request is opened or master is pushed to. There are also [methods that will prevent it from triggering](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-filters) when changes are made only to documentation or data, etc. https://docs.github.com/en/actions/use-cases-and-examples/creating-an-example-workflow

Additionally, although you already test using an Ubuntu container, with a workflow you can test on other distros if you decide to do so in the future. One way to do that is by using the *container* key: https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/running-jobs-in-a-container

PR #243, opened 2025-02-13
M1 macOS build compatibility fixes

Jorgeelalto:
This is only a draft of the stuff necessary to start building this game for M1 macOS, I don't know if it will even be possible at all.

The dependencies are able to be installed with `brew` no problem. First issue comes when generating the build files (using `cmake -B build`), which explains the fix in line 35 of the `CMakeLists.txt` file. In my case I didn't specify `CMAKE_BUILD_TYPE` and the file couldn't handle that case so I added that code to override it.

Next are the includes in macOS, which, if installing the dependencies with `brew` (the usual thing to do) requires including headers and libraries from a specific location. I know this `/opt/homebrew` location is specific for ARM Macs too (Intel is a different folder). I am not sure how to streamline this as I couldn't find any other `include_directories()` or `link_directories()` directives in the file.

Last is the hack in `src/lincity-ng/lc_error.h` which allows me to not include `error.h` which is unavailable in macOS. Unfortunately it is recognized as an UNIX system (is that something you can control or is it a CMake default?) so I had to bypass it. Again, not sure how to proceed with that.

Let me know if you want to proceed with this. I have some free time to try to build lincity-ng in my Mac but I understand if this is something you might not wanna pursue due to my lack of experience with CMake.


Jorgeelalto:
Worth mentioning, after these hacks the program finishes compilation but fails at linking:

```
[ 99%] Linking CXX executable ../../bin/lincity-ng
ld: warning: ignoring duplicate libraries: '../../external/physfsrwops/libphysfsrwops.a', '../PhysfsStream/libphysfsstream.a', '../gui/liblincity_gui.a', '../lincity/liblincity_lib.a', '../lincity/modules/liblincity_lib_modules.a', '../tinygettext/libtinygettext.a'
Undefined symbols for architecture arm64:
  "_CFBundleCopyBundleURL", referenced from:
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
  "_CFBundleGetMainBundle", referenced from:
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
  "_CFRelease", referenced from:
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
  "_CFStringGetCString", referenced from:
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
  "_CFURLCopyFileSystemPath", referenced from:
      getBundleSharePath(char*) in getBundleSharePath.cpp.o
  "_error_at_line", referenced from:
      Dialog::saveGameStats() in Dialog.cpp.o
      initPhysfs(char const*) in main.cpp.o
      initPhysfs(char const*) in main.cpp.o
      initialize_monthgraph() in ScreenInterface.cpp.o
      initialize_monthgraph() in ScreenInterface.cpp.o
      initialize_monthgraph() in ScreenInterface.cpp.o
      initialize_monthgraph() in ScreenInterface.cpp.o
      ...
ld: symbol(s) not found for architecture arm64
c++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/lincity-ng] Error 1
make[1]: *** [src/lincity-ng/CMakeFiles/lincity-ng.dir/all] Error 2
make: *** [all] Error 2
```

This is the full log: [log.txt](https://github.com/user-attachments/files/18786490/log.txt)

dbear496:
I am definitely interested in better support for Macs.

> First issue comes when generating the build files (using cmake -B build), which explains the fix in line 35 of the CMakeLists.txt file. In my case I didn't specify CMAKE_BUILD_TYPE and the file couldn't handle that case so I added that code to override it.

On Linux and Windows, `CMAKE_BUILD_TYPE` is set to `Release` by default. However, the [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) states that the default _may_ be an empty string, which would not be desireable. So setting an explicit default in CMakeLists.txt should be the correct thing to do.

> Next are the includes in macOS, which, if installing the dependencies with brew (the usual thing to do) requires including headers and libraries from a specific location. I know this /opt/homebrew location is specific for ARM Macs too (Intel is a different folder). I am not sure how to streamline this as I couldn't find any other include_directories() or link_directories() directives in the file.

The include/link directories should ideally be set by a cmake configuration file that is installed to your system by the libraries themselves. The locations that CMake searches for this configuration file is detailed [here](https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure). This configuration file should give the build system the correct include directories and link library locations. If you find that these configuration files are not installed where they should be, you may choose to look online for some cmake 'find module's that you can copy into the correct locations and adapt to your needs.

> Last is the hack in src/lincity-ng/lc_error.h which allows me to not include error.h which is unavailable in macOS. Unfortunately it is recognized as an UNIX system (is that something you can control or is it a CMake default?) so I had to bypass it. Again, not sure how to proceed with that.

CMake sets the `UNIX` variable for UNIX and UNIX-like systems, and evidently apple is included in that. We could use `CMAKE_SYSTEM_NAME` instead. Or we may use
```
#if defined(UNIX) && ! defined(APPLE)
```
I'm not sure at the moment which will be better, but I'm leaning toward the latter since it requires the least modification. Also, the lc_error.h header is currently used in only one place, so it may be better to just remove that header altogether. After all, the error.h functionality is not very C++-ish, and we could use exceptions instead.

As for the link errors, I'm pretty sure the `getBundleSharePath` logic is specifically for Apple systems...so if it doesn't work on Apple, then that kinda defeats the purpose. Also `getBundleSharePath` appears to be totally unused -- perhaps it was left over from version 2.9.0. So I expect that entire file may be removed along with the `extern` declaration in main.cpp.

CMakeLists.txt
```diff
@@ -28,7 +32,9 @@ set(CMAKE_FIND_PACKAGE_PREFER_CONFIG TRUE)
 
 ### build configurations
 
-string(TOUPPER ${CMAKE_BUILD_TYPE} BUILD_TYPE_UPPER)
+if(CMAKE_BUILD_TYPE)
```
dbear496, line 35: It may be prudent to set the `CMAKE_BUILD_TYPE` to `Release` if it is empty. The CMake docs state, "The default value is often an empty string, but this is usually not desirable and one of the other standard build types is usually more appropriate."

CMakeLists.txt
```diff
@@ -19,6 +19,10 @@ include(CheckIncludeFiles)
 include(GNUInstallDirs)
 include(GetGitRevisionDescription)
 
+include_directories(/opt/homebrew/include)
```
dbear496, line 22: The include/link directories *should* be configured by find_package. If this is not so, it may be an issue with the package configure modules supplied by your system.

dbear496:
Since Mac doesn't provide `error.h`, it is very likely that you need the definition of `error_at_line` from `lc_error.c`, so try mangling or removing the `#ifndef` directive in `lc_error.c`. I should have noted this when you changed the `#ifdef` directive in `lc_error.h`, but it didn't occur to me.

Now I'm just a tad worried about the extra underscore in your error log (`error_at_line` vs. `_error_at_line`). There is a small chance that one of the header files on Mac has a `#define error_at_line _error_at_line` but then doesn't define `_error_at_line`, which we would have to look into if that ends up being the case. But perhaps this is just from g++ name mangling, which is benign.

Jorgeelalto:
Well well well... I got it to compile 😃

I ended up defining an `error()` macro and pointing `lc_error()` to it. It is now building and running!

Screenshot 2025-04-16 at 11 26 30

I will try to make a proper PR with this information. I think getting a clean lc_error.h for macOS will be relatively straightforward (using the `APPLE` constant as you mentioned). Then I will try to get my hands on the CMake files.

dbear496:
That's great news! making an `error` macro for Mac should be pretty straight forward. I suspect you will be able to drop in the implementation of `error_at_line` that is already implemented in `lc_error.c`, and this will only require changing the `#ifdef`s in `lc_error.c` and `lc_error.h` to use the `APPLE` identifier. `getBundleSharePath.cpp` can be deleted entirely, so that should also be straight forward. The only difficulty might be getting CMake find modules to play nice.

thomashrabe:
@Jorgeelalto I got the game to compile on my Intel Mac with almost identical changes as yours. However, it segfaults when I run it. Maybe we can figure the CMake stuff together? I can not push my branch as I do not have the rights and similarly can not download yours.

dbear496:
I pushed a commit to your repo that cleans up some of the hackiness and makes a `error_at_line` implementation available on Mac. Please see if this works.

I did not address the issue with the include/link directories at all because this will require troubleshooting on a Mac.

Jorgeelalto:
Hi @dbear496, I can confirm it still builds and runs properly with your commit, thanks a lot.

@thomashrabe not sure why you wouldn't be able to clone my fork, it's public...

thomashrabe:
@Jorgeelalto Thanks, I didnt realize you had the fork public, I got that stuff now. The segfaults happen because the default installation dir is set to `set(CMAKE_INSTALL_PREFIX "/usr/local")` . I'm running in debug mode with `--app-data ${workspaceFolder}/build/share/lincity-ng` (VSCode) and it still tries to find files in the default install dirs. Thats where its failing. I did not install the package.

dbear496:
Most of these changes are effectively in version 2.14.0. The only change that is not in 2.14.0 is the hard-coded include/link directories, which is a hack fix for malformed or missing cmake config modules on the target system. So I am closing this PR.

PR #254, opened 2025-05-13
HW acceleration and blitting in SDL video mode

### LINKED ISSUE #202: high CPU usage during game ?
(Splitted from #201.)

During the game (for example after loading Rocket_98) CPU usage is at about 65% (max zoom in) to 110% (max zoom out).

CPU is Intel Core i7-7500U @ 2.70GHz.

dbear496:
This PR implements hardware acceleration in SDL video mode and blits the gameview and minimap in hardware. This dramatically improves performance for SDL video mode: hardware acceleration alone puts it on par with GL mode, and blitting improves performance even further.

Previously, SDL mode was very slow because textures were primarily stored in software, and they were transferred to the GPU on every use every frame -- including duplicate transfers when the same texture was used multiple times in the same frame.

GL mode is disabled because it does not support blitting (and the performance improvement of SDL mode makes it obsolete anyway).

OpenGL and SDL_gfx library dependencies are removed. And minimum SDL2 version is increased to 2.0.18.

Fixes #202


dbear496:
> In 2.12.0 SDL mode looks very pixelated (while GL is OK). Tested on Linux, just starts with `lincity-ng -s` and the background image looks bad. Is this also addressed by these patches?

This isn't specifically addressed. However, moving all the texture storage from software to hardware may have fixed the issue. Also, this could be an issue with Direct3D. It would be helpful if you could test with the latest version and share screenshots. I tested the latest Windows version in Wine, but it didn't appear pixelated.

I am aware that the background shows a Moire pattern when zoomed out. And it is also not terribly efficient because it loops through each background tile. This can be fixed by using a background image (which doesn't zoom or move) instead of tiles. But this is non-critical as it's only a problem when zoomed pretty far out, so I decided not to address it in this PR.

dbear496:
Okay, I see now. (Thanks for the screenshots.) This looks to me like SDL mode isn't using anti-aliasing, so the menu images look pixelated when they are scaled up for larger screens. The pixelization isn't there if the window is shrunk to a smaller size. The pixelization isn't noticeable in-game unless you zoom in really far.

There are two things that should both be done about this: create larger background images for the menu screens, and add a configuration option for anti-aliasing.

v1993:
The term you're looking for is bilinear filtering; anti-aliasing is a different concept altogether. It should be possibly to enable it by doing

```c
SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "1")
```

somewhere during initialization (only once and before any textures are created).

dbear496:
Yes, you're right. I had that wrong. It is the texture filter, not anti-aliasing.

dbear496:
I tried setting `SDL_HINT_RENDER_SCALE_QUALITY` to 'nearest', and the background image in the main menu is definitely less pixelated. However, it also causes grid lines on the map, which I don't prefer.

v1993:
"Nearest" would correspond to doing no filtering, which is supposedly the default behavior, so I'm surprised it made any difference. Either way, it's also possible to set filtering mode per-texture with `SDL_SetTextureScaleMode`.

dbear496:
Okay, those latest two commits should fix the pixelization in the menu backgrounds.

dbear496:
Yes, anisotropic filtering with SDL has a result similar to GL even with the smaller image size. The larger image size gives better results for 2K screens.

I recompressed the images with a better compression level and force-pushed.

PR #278, opened 2025-07-11
fix strange texture sampling by rounding texture translation

### LINKED ISSUE #276: Artifacting on some icons in the UI
Noticed this after firing up the game to take some screenshots. This might be affecting some other UI elements, like minimap mode switch buttons, but the most obvious case is the "accept" button here:

Image

It looks very strange - the top half is nice and crispy, but then something goes sideways and everything gets pixelated. I don't think this was an issue on the OpenGL renderer.

dbear496:
It seems the GPU cannot be trusted with rounding floats with a fractional part close to 0.5. This PR rounds the absolute position of textures when they are not being stretched.

fixes #276

### DIFF
--- a/CHANGELOG.d/fix-image-sampling.md
+++ b/CHANGELOG.d/fix-image-sampling.md
@@ -3,7 +3,7 @@
 ###### Gameplay
 
 ###### User Interface
-- Fixed an issue causing some UI components to appear blurry.
+- Fixed an issue causing some UI components to appear distorted.
 
 ###### Internal
 
--- a/CHANGELOG.d/fix-image-sampling.md
+++ b/CHANGELOG.d/fix-image-sampling.md
@@ -1,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+- Fixed an issue causing some UI components to appear blurry.
+
+###### Internal
+
+###### Documentation / Translation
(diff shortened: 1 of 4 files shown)


v1993:
This is definitely much better, although is there any particular reason to be using NN filtering there instead of linear?

dbear496:
> This is definitely much better, although is there any particular reason to be using NN filtering there instead of linear?

Nearest filtering produces a more crisp result when scaling/stretching isn't an issue.

dbear496:
Rounding the position of an unscaled and unstretched image in software rather than hardware when using nearest filtering technically shouldn't change the result at all -- it just guarantees that every pixel sample is rounded the same direction.

v1993:
If being unscaled means lack of uniform scaling (as it seems), then yeah, NN seems like the right tool here. I wonder if there's some underlying cause leading to images consistently ending up at a 0.5 pixel offset and if this is just going to hide that, though.

CHANGELOG.d/fix-image-sampling.md
```diff
@@ -0,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+- Fixed an issue causing some UI components to appear blurry.
```
v1993, line 6:
```suggestion
- Fixed an issue causing some UI components to appear distorted.
```

I don't think these artifacts could be described as "blur", as everything looks crisp, just jagged in odd ways.

v1993 (commented):
It's probably a good idea to take new screenshots once this is merged. No rush for that, though, especially since screenshots are fetched from the web rather than the release tarball.

v1993 (approved):

PR #279, opened 2025-07-12
CMake option tweaks and additions

v1993:
* Add an option to override the version suffix as discussed in #277.
* Rename `DISABLE_GL_MODE` to `LINCITYNG_DISABLE_GL_MODE` on CMake side so that LinCity-NG specific options have a common prefix.

### DIFF
--- a/CHANGELOG.d/version-suffix-override.md
+++ b/CHANGELOG.d/version-suffix-override.md
@@ -1,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+
+###### Internal
+- Added a CMake option `LINCITYNG_VERSION_SUFFIX` to allow overriding the version suffix. This is mostly useful for packaging, especially when building from a source tarball.
+
+###### Documentation / Translation
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,11 +79,17 @@ find_program(INCLUDE_WHAT_YOU_USE NAMES include-what-you-use)
 
 
 ### compute the LinCity-NG version
-git_describe_working_tree(FULL_PROJECT_VERSION --match lincity-ng-*)
-if(FULL_PROJECT_VERSION)
-	string(SUBSTRING ${FULL_PROJECT_VERSION} 11 -1 FULL_PROJECT_VERSION)
+set(LINCITYNG_VERSION_SUFFIX "" CACHE STRING "Custom version suffix to use; mostly useful for downstream packaging.")
+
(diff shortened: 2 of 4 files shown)


CHANGELOG.d/version-suffix-override.md
```diff
@@ -0,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+
+###### Internal
+- Add a CMake option `LINCITY_VERSION_SUFFIX_OVERRIDE` to allow overriding the version suffix. This is mostly useful for packaging, especially when building from a source tarball.
```
dbear496, line 8: This needs to be past tense: "Added CMake option ...".

CMakeLists.txt
```diff
@@ -79,11 +79,17 @@ find_program(INCLUDE_WHAT_YOU_USE NAMES include-what-you-use)
 
 
 ### compute the LinCity-NG version
-git_describe_working_tree(FULL_PROJECT_VERSION --match lincity-ng-*)
-if(FULL_PROJECT_VERSION)
-	string(SUBSTRING ${FULL_PROJECT_VERSION} 11 -1 FULL_PROJECT_VERSION)
+set(LINCITY_VERSION_SUFFIX_OVERRIDE "" CACHE STRING "Custom version suffix to use; mostly useful for downstream packaging.")
```
dbear496, line 82: I prefer we omit the '_OVERRIDE' part of the name. `LINCITYNG_VERSION_SUFFIX` should do.

CMakeLists.txt
```diff
@@ -79,11 +79,17 @@ find_program(INCLUDE_WHAT_YOU_USE NAMES include-what-you-use)
...
+
+if (NOT LINCITY_VERSION_SUFFIX_OVERRIDE)
```
dbear496, line 84: Use `NOT DEFINED LINCITYNG_VERSION_SUFFIX OR LINCITYNG_VERSION_SUFFIX STREQUAL ""` so a falsy suffix can be used. The `NOT DEFINED` part maybe can be omitted since we `set` right before, but we need to be careful of the `-U` cmake option.

I'm in favor of inverting this if statement.

dbear496 (changes requested):

CMakeLists.txt
v1993: > I'm in favor of inverting this if statement.

I personally prefer `else` blocks to be the shorter of the two when there's a significant length difference, but sure.

PR #282, opened 2025-07-12
Use presice scrolling where applicable

v1993:
This works really nicely with my touchpad. It was added in SDL2 2.18.0, so no need for a version check.

For now, a complete replacement isn't done as other places rely on it working as an integer accumulator - FWIW, SDL3 dropped that, so accumulation will have to be done manually when migrating there.

### DIFF
--- a/CHANGELOG.d/precise-scrolling.md
+++ b/CHANGELOG.d/precise-scrolling.md
@@ -1,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+- Added smooth scrolling support for devices where it is available, e.g. touchpads.
+
+###### Internal
+
+###### Documentation / Translation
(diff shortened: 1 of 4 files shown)


CHANGELOG.d/precise-scrolling.md
```diff
@@ -0,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+- Support smoother scrolling on devices where it is available, e.g. touchpads.
```
dbear496, line 6: Please make this past tense and preferably start with the word "Added", e.g. "Added smooth scrolling ..."

dbear496 (changes requested):
One nitpick with the changelog entry. The rest looks good, and the smooth scrolling has a nice feel.

dbear496 (approved):

PR #283, opened 2025-07-13
Smoothness improvements, especially on VRR displays

### LINKED ISSUE #281: Issues with LinCity-NG's handling of frame output
It seems like LinCity-NG tries to avoid redrawing frames if possible - this is quite evident in menus that will not draw at all when there's nothing changing. A lot of the time it's not a problem with lincity-ng's low-framerate pre-rendered animations and is probably great for power saving, especially with VRR, but it does result in some issues:

* Mouse movement looks quite choppy if you have VRR enabled, both in-game and in menus; my guess is that it ends up getting locked to the bottom boundary of the VRR range, which would be 60 out of potential 144 Hz for me
* Moving around with arrow keys or scrolling in documentation by holding down a button is done at 40 Hz, which is far from smooth
* Moving around or scrolling in documentation by dragging allows framerate to rise uncontrollably, potentially beyond that of the screen, but is admittedly relatively smooth

I think the solution here would be to redraw as fast as possible when there's mouse movement or scrolling of any kind taking place - this is already done for dragging around and scrolling with mouse, but should also be performed for general mouse movement (as with VRR screen only updates when the game draws or bottom boundary of VRR times out, whichever is earlier) and scrolling/moving around with keyboard. Additionally, enable VSync to ensure that "as fast as possible" is not actually faster than the screen can show.

v1993:
Fixes #281. With these changes LinCity-NG feels a lot nicer to play on my 144Hz VRR monitor!

More specific outline of changes:

* Enable VSync - only done for the SDL painter as the OpenGL one is apparently being phased out
* Redraw on mouse motion to ensure that pointer does not feel choppy on a VRR screen
* Introduce a `Component::requestFastNextFrame` method that basically allows components to ask the main loop to render next frame as early as possible
* Run a GUI update if we're drawing anyways

### The tricky stuff

A known ~~issue~~ *peculiar behavior* with this PR is that when moving the mouse the simulation somewhat slows down (even if it stays quite fast) on the fastest speed, as it gets limited to a single simulation frame per render frame; this previously already happened with things that triggered redrawing from the event polling loop, like dragging the map, but is a lot more obvious now, especially with render frames themselves also being capped by Vsync. I don't think it's really a *huge* problem, but there are at least few ways out of it I can think of:

#### Hard way out: Move simulation to another thread

The theoretically best solution, but also obviously the hardest one. I guess you'd want some kind of mailbox synchronization where the main thread grabs the last available state and displays it. Extremely out-of-scope.

#### Naive way out: cap the top speed and allow multiple simulation frames per frame

Much easier to code (I mean, I already made an implementation), but poses the opposite problem: now render frames might get slowed down by the simulation code. For example, rocket 98 can run about 500 simulation steps per second on my laptop, which means I have to set `SIM_DELAY_FAST` to `2` to avoid major slowdown, but this won't be constant across different hardware and different cities (an early game city runs at basically lightspeed), meaning it's not a great solution.

#### Seemingly reasonable way out?: use a high-resolution timer to find out how many simulation ticks can fit in a render frame

You can try to accurately measure how long a frame takes and try to compute for how long you can run physics frames alongside it. The problem I've ran into is that presentation time seemingly doesn't really correspond to how much free time you have left over and simulation isn't the last thing to run before presentation either. I've made a somewhat functioning prototype, but it's still prone to weird fluctuations and needs a safeguard to avoid spiraling into single-digit framerates.

Honestly? At this point I'd be happy to just let things get capped by vsync. Don't move your mouse if you want the game to go truly brrr.

P.S.: @dbear496 I remember you've mentioned reworking the main loop? If so, I guess some of these changes will conflict with that, but I think they're at least worth discussing. Also, I wonder what do you think about going the route most games handle the main loop and running it at a fixed rate at all times?

### DIFF
--- a/src/gui/Component.cpp
+++ b/src/gui/Component.cpp
@@ -204,6 +204,12 @@ Component::setDirty(const Rect2D& rect)
         parent->setChildDirty(this, rect);
 }
 
+void
+Component::requestFastNextFrame() {
+    if(parent)
+        parent->requestFastNextFrame();
+}
+
 Child&
 Component::findChild(Component* component)
 {
--- a/src/gui/Component.hpp
+++ b/src/gui/Component.hpp
@@ -151,6 +151,13 @@ protected:
     virtual void setDirty(const Rect2D& area);
 
     /**
+     * Schedule next frame to happen early and be a GUI frame.
+     * Useful for smooth animations that are not directly tied to an event,
+     * such as scrolling by holding a button.
+     */
+    virtual void requestFastNextFrame();
+
+    /**
      * Used to parse attributes (from an xml stream for example). Currently
      * parses only the name attribute.
      * @return True if the attribute has been used, false else.
--- a/src/gui/Desktop.cpp
+++ b/src/gui/Desktop.cpp
@@ -104,7 +104,8 @@ Desktop::draw(Painter& painter)
             SDL_SetCursor(cursor);
     }
     dirtyRectangles.clear();
-    force_redraw = false;
+    force_redraw = force_redraw_next_frame;
+    force_redraw_next_frame = false;
 }
 
 bool
@@ -220,6 +221,11 @@ Desktop::setDirty(const Rect2D& rect)
     dirtyRectangles.push_back(rect);
 
     Component::setDirty(rect);
+}
+
+void
+Desktop::requestFastNextFrame() {
+    force_redraw_next_frame = true;
 }
 
 /** @file gui/Desktop.cpp */
--- a/src/gui/Desktop.cpp
+++ b/src/gui/Desktop.cpp
@@ -32,6 +32,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #include <string>               // for char_traits, basic_string, operator<<
 
 #include "Child.hpp"            // for Childs, Child
+#include "Event.hpp"            // for Event
 #include "ComponentLoader.hpp"  // for createComponent
 #include "Style.hpp"            // for parseStyleDef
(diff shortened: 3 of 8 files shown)


dbear496:
> A known ~issue~ _peculiar behavior_ with this PR is that when moving the mouse the simulation somewhat slows down

I consider this an _issue_. It is unacceptable for simulation speed to dip just from moving the mouse. I already fixed this in #254. and reintroducing it is a step backward.

> Enable VSync

This unnecessarily slows simulation. Of your proposed solutions, the hard way is the best, but of course.....it's hard -- bordering on infeasible. The other two proposed solutions are a big no. The naive way is going back to version 2.9 behavior. The "seemingly reasonable" way is a hack that tries to predict the timing of the host system in a way that bypasses API.

I propose splitting the display update to a new thread. Simulation, animation, rendering, event processing, and GUI updates all stay in the main thread. This isolates the blocking behavior of the display update from the rest of the program and doesn't complicate data synchronization between simulation and rendering.

> Redraw on mouse motion to ensure that pointer does not feel choppy on a VRR screen

The cursor should be managed by the host system, and I don't think it is the game's responsibility to ensure it is smooth. So it seems to me this is a workaround for an issue with the host system.

> Introduce a Component::requestFastNextFrame method that basically allows components to ask the main loop to render next frame as early as possible

As I mentioned in the review comment, I think this function is doing way more than is actually needed.

> Run a GUI update if we're drawing anyways

This seems reasonable to me.

> I remember you've mentioned reworking the main loop?

I reworked the `main` _function_ -- not the main loop. You can see the changes in the _init_ branch on my fork.

dbear496 (changes requested):

PR #270, opened 2025-06-28
Clean up unused files

dbear496:

PR #284, opened 2025-07-13
Use bona fide gettext

### LINKED ISSUE #274: Build fails when using Ninja cmake backend
A curious error I've encountered when trying to build LinCity-NG:

```
> cmake -B build -DCMAKE_BUILD_TYPE=Release -GNinja
...
> cmake --build build --parallel
ninja: error: build.ninja:18329: multiple rules generate share/lincity-ng/locale/messages.pot
```

With ninja being the generally preferred "second-level" build system due faster build times compared to the default `make` it'd be nice to have this issue fixed.

This is probably caused by having both a `messages.pot` file and a custom target named the same in `data/locale/CMakeLists.txt`.

dbear496:
Removes the custom gettext implementation and instead uses official gettext/libintl.

- Translation locale files are installed in binary (.mo) format, converted from textual (.po) format at build time.
- Combines the translation files from GUI XML and source code into a single file. The PO template files are also combined into one.
- Adds documentation on build-time configuration and run-time configuration.
- Moves the dependency list out of the readme because it was taking too much space there.
- Fixes #274

depends on #270 
depends on #306


v1993:
1. Any reason to be moving unused code files to `contrib` instead of just deleting them? tinygettext in particular is a publicly available library anyways, and these files will persist in Git history one way or another.
2. `ENABLE_NLS`/`LINGUAS` CMake options probably should have a `LINCITYNG_` prefix.
3. It's likely a good idea to make a header for common shorthand macros like `_`,`N_`, etc. instead of defining them in each source file where they are used, like `lincity_gettext.h`. DRY. They aren't even strictly identical right now - sometimes it's `#define N_(MSG) gettext_noop(MSG)`, sometimes it's `#define N_(MSG) MSG` - although it makes no functional difference, since both resolve to `MSG`, it's still an inconsistency.
4. I was surprised to see that Gettext really expects you to have `gettext.h` as part of your project if you aren't using `libintl` directly. Quite a choice on their part.

v1993:
1. Makes sense.
2. I was under impression that these are not evironment variables, but I can't verify rn since gnu.org is down - talk about bad timing. Either way, the option `--disable-nls` is specific to autotools and cmake will not let you add it in that form. CMake's `option` is for adding options that are set with `-DNAME=VALUE`. To be honest, I think it would be alright for lincity-ng to not handle the case of missing libintl due to how it's effectively omnipresent on environments lincity-ng is expected to run on (which would also allow to avoid gettext.h altogether), but you might disagree.
3. Yeah, the idea is to put them in a header, but avoid including that from other header files.

v1993:
> I believe `LINGUAS` is an environment variable and not `ENABLE_NLS`

Seems to be right, yeah.

> Regardless, these are variables the user will likely want to apply uniformly to an entire application, so a separate variable for each individual library is not desirable.

This would require every program in build to adhere to the same variable naming convention, though, and I don't think `ENABLE_NLS` is standard in CMake in any way.

> I know libintl was not present in my mingw toolchain until I explicitly installed it earlier this week, so maybe not omnipresnet.

Fair, then I think a warning should be issued if user doesn't explicitly disable it first, as they might be left confused why are there no translations otherwise. Maybe a better way to put it would be that no user of a lincity-ng is likely to *want* to build without gettext and its absence is most surely a mistake.

> And the gettext manual says disabling NLS (`ENABLE_NLS` / `--disable-nls`) should remove dependency on libintl and/or other NLS dependencies.

Gettext manual is a bit of an odd beast, to be honest. It only gets to using `gettext.h` and extras it offers very late into the manual and otherwise talks about including `libintl.h` directly. But, ultimately, it really is up to you if this functionality is desired.

dbear496:
I fixed the XML parsing so it is all locale-_independent_ now. While I was doing that, I also updated the GUI XML parsing to use libxml++ instead of libxml2.

I also merged in PR #285 because it had fallen so far behind, and it was weirdly interdependent with this PR.

Some things still left to do in this PR:
- Call `SDL_SetHint(SDL_HINT_APP_NAME, PRETTY_NAME)`.
- replace `SDL_INIT_EVERYTHING` with a more specific bitmask.
- Remove unnecessary checks around `SDL_Quit` and `TTF_Quit`.
- Issue a warning when NLS is disabled due to not finding libintl.
- Change the appearance of the language selector in the config menu when NLS is disabled.

PR #285, opened 2025-07-13
Clean up the program initialization code

dbear496:
Cleans up the `main` function and some of the `Config` logic.

depends on #284


dbear496:
In the long run, I'd like `sound` to not be a singleton -- instead pass it as a resource to modules that need it.

PR #286, opened 2025-07-18
Optimize images

v1993:
Does not touch images that are not installed to avoid inflating git history and thus cloning time for no final size benefit.

Optimized using [`ect`](https://github.com/fhanau/Efficient-Compression-Tool).


data/images/gui/buttonpanel/pottery.png
dbear496, line 1: This image is unused. It should go to contrib.

data/images/gui/panel.png
dbear496, line 1: move to contrib

dbear496 (approved):
A few of these files are actually unused, so they should be moved to contrib and don't need optimization.

Wow! I didn't realize how disorganized the images folder is -- so many things in the wrong places. Re-organizing can be done in a separate PR.

v1993:
> I used oxipng -o max in the past, is ect better?

It seems to generally outdo other options from what I've seen in comparisons.

> There are also a few other unused images, I had a list, will post it end of month if you want to avoid recompressing them.

I don't think it's critical if it's just a few images, but it'd still be nice to at least move them to `contrib` for sure.

CHANGELOG.d/optimize-images.md
```diff
@@ -0,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+
+###### Internal
+- Optimize size of image assets.
```
dbear496, line 8: This should be past-tense.
```suggestion
- Optimized size of image assets.
```

dbear496 (changes requested):

PR #290, opened 2025-08-02
remove spaces in files

FabioPedretti:
spaces are replaced with _ completing previous commits, e.g. 443895921fdb8


dbear496:
If the motivation here is to match 4438959, then the music files should be renamed in a similar manner to 4438959. I.e. with the same use of dashes and underscores.

dbear496:
I think option 1 is the best way to go for these texture files unless the spaces are actively causing issues. Option 3 is a good way to confuse that person in the future.

PR #308, opened 2025-09-05
release notes

dbear496:
Adds some notes on how to publish a release.

@v1993 Could you add to this file the list of steps for publishing the flatpak/flathub release?

Depends on #284

### DIFF
--- a/doc/RELEASE.md
+++ b/doc/RELEASE.md
@@ -40,6 +40,8 @@
     - build/linux/lincity-ng-x.x.x-Source.tar.xz.sha256
     - build/win64/lincity-ng-x.x.x-win64.zip
     - build/win64/lincity-ng-x.x.x-win64.zip.sha256
+    - build/win64/lincity-ng-x.x.x-win64.exe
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -55,6 +56,22 @@ When the game is installed, you may run it with:
...
+
+As an alternative to building from source, there are several places to get
+pre-built binary packages:
+
+- Linux distribution repositories. This is likely the best way to install
+  LinCity-NG pre-built binaries for Linux users. However, LTS and stable
+  distributions, by their very nature, may not offer the latest version of
+  LinCity-NG.
+- [Flathub](https://flathub.org/apps/io.github.lincity_ng.lincity-ng). As of
+  version 2.14.1, LinCity-NG is available on Flathub.
```
v1993, line 69: Probably a good idea to add a short outline fo pros and cons here like it's done for other options? I.e.: likely to be more recent than a distro package on stable distributions, but probably bigger than other options due to nature of flatpak.

PR #320, opened 2025-09-23
use $version whereever possible in RELEASE.md

dbear496:


dbear496:
> Proposed addition: use `xmlstarlet` to automate insertion of the release tag into metainfo

#321 was caused solely because I did not know that Flathub was particular about the order of the `release` tags. So a script would not have prevented that mistake. For the time being, I want to keep the entire release process documented in the RELEASE.md file with only a couple lines per step (i.e. the starlet command is too long to be included in RELEASE.md).

I have in the back of my mind to eventually automate the release process, and when that time comes, I'll probably add the tag to the metainfo file with xsltproc. I'd like to use xsltproc as opposed to xmlstarlet because other parts of the project already depend on xsltproc.

I don't want to automate the release process yet. Since the RELEASE.md documentation is still very new, I want to give some time to allow the process to be fully fleshed out before solidifying it into a script.

v1993:
I see. I think that using `xsltproc` for this purpose will run into the problem of being quite annoying as you'd have to generate the XSLT file with version/date values in it (it can't read them from the environment); at this point you might as well use a slightly hacky but extremely portable solution with `sed` or a similar "dumb" text manipulation tool to insert a new entry right after `<releases>`.

> For the time being, I want to keep the entire release process documented in the RELEASE.md file with only a couple lines per step (i.e. the starlet command is too long to be included in RELEASE.md).

Well, you *can* make it into a one-liner if you must :) .

PR #327, opened 2025-09-26
Use xsltproc for adding a release entry

v1993:
Continuation of my idea in #320, implemented with `xsltproc`. The command is quite long, but that's mostly due to long (and repeating) file names.

### DIFF
--- a/contrib/add-release.xsl
+++ b/contrib/add-release.xsl
@@ -6,6 +6,15 @@
   <xsl:param name="version"/>
   <xsl:param name="date"/>
 
+  <xsl:template match="/">
+    <xsl:if test="not($version) or not($date)">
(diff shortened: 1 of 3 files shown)


dbear496:
Is it possible to detect if the correct release tag is already present and not add a duplicate?

v1993:
> Did you verify that xsltproc supports reading and writing the the same file? Sometimes this will work for shorter files but fail for larger files when an IO buffer fills up.

I've checked the source code and the file is parsed before any processing begins. To be honest, in case of streaming it would be rather odd to only open (and thus erase) the file after being done with the first segment.

> Is it possible to detect if the correct release tag is already present and not add a duplicate?

Good idea, added that check along with throwing an error in case of missing parameters.

dbear496 (approved):

PR #343, opened 2026-05-02
Initial SDL3 port. This runs natively with SDL3, not using sdl2-compat.

### LINKED ISSUE #340: libsdl2 deprecation and libsdl3
Debian is going to replace libsdl2 with sdl2-compat (compatibility layer for libsdl3): https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1130484

Arch and Fedora have already made this switch.

In the longer term, lincity-ng should probably directly supports sdl3 for best integration.

myoung008:
This is a very basic port to SDL3.  There may be newer better ways to use parts of the SDL api, but the goal of this is just to move things to the new version as simply as possible.  I did not use any kind of AI for this patch.  I used the header renaming script that is a part of SDL3 to make the header changes across the whole project.

fixes #340


mk/cmake/modules/SDL2/FindSDL2.cmake
```diff
@@ -143,7 +143,7 @@ environment CMAKE_INCLUDE_PATH to modify the search paths.
 
 Note that the header path has changed from SDL/SDL.h to just SDL.h
 This needed to change because "proper" SDL convention is #include
-"SDL.h", not <SDL/SDL.h>.  This is done for portability reasons
```
v1993, line 146: This is in a `FindSDL2.cmake` file, so this change isn't appropriate. Either leave the file untouched or delete it and the rest of SDL2 modules if they're not used anymore. Are SDL3 finders needed?

src/gui/PainterSDL/PainterSDL.cpp
```diff
@@ -274,21 +275,24 @@ PainterSDL::pushRenderTarget(Texture *target) {
   assert(t);
   #ifndef NDEBUG
   {
-    int tflags;
-    HANDLE_ERR(SDL_QueryTexture(t->tx, NULL, &tflags, NULL, NULL));
+    ;
```
v1993:
```suggestion
```

src/gui/PainterSDL/TextureManagerSDL.cpp
```diff
@@ -22,7 +22,7 @@
 
 #include "TextureManagerSDL.hpp"
 
-#include <SDL.h>           // for SDL_ConvertSurfaceFormat, SDL_FreeSurface
+#include <SDL3/SDL.h>      // for SDL_ConvertSurface
```
v1993: But there are no actual replacements in the file itself?

src/gui/PainterSDL/TextureSDL.cpp
```diff
@@ -22,16 +22,19 @@
...
 #include <iostream>
 
 TextureSDL::TextureSDL(SDL_Texture *tx) : tx(tx) {
   assert(tx);
-  if(SDL_QueryTexture(tx, NULL, NULL, &width, &height))
+  float w, h;
+  if( !SDL_GetTextureSize(tx, &w, &h))
     throw std::runtime_error(SDL_GetError());
+  width=w;
+  height=h;
```
v1993: I think you can just directly access `tx->w` and `tx->h` instead of calling `SDL_GetTextureSize` and dealing with floats. I guess `width`/`height` could be removed entirely in the future, but a more direct assignment should do for now.

src/gui/Event.cpp
```diff
@@ -18,60 +18,58 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
...
             mousebutton = event.button.button;
             break;
-        case SDL_MOUSEWHEEL:
+        case SDL_EVENT_MOUSE_WHEEL:
             type = MOUSEWHEEL;
-            scrolly = event.wheel.y;
-            scrolly_precise = event.wheel.preciseY;
+            scrolly = event.wheel.integer_y;
+            scrolly_precise = event.wheel.y;
             #if SDL_VERSION_ATLEAST(2,26,0)
```
v1993: Always true in SDL3 and thus should be removed alongside any other SDL version checks.

src/gui/Event.hpp
```diff
@@ -74,8 +74,13 @@ class Event
     int mousebutton;
     /// mouse button state (can be decoded with SDL_BUTTON macros)
     Uint32 mousebuttonstate;
+
     /// symbol of the key that has been pressed (see SDL_keysym)
-    SDL_Keysym keysym;
+    //SDL_Keysym keysym;
```
v1993: Outdated comment + commented out code.

src/gui/Gradient.cpp
```diff
@@ -86,19 +86,17 @@ Gradient::resize(float width, float height)
...
-    SDL_Surface* surface = SDL_CreateRGBSurface(SDL_SWSURFACE,
-                                                (int) width, (int) height,
-                                                32, 0x000000ff,
+    SDL_Surface* surface = SDL_CreateSurface(  (int) width, (int) height,
+                                                SDL_GetPixelFormatForMasks(32, 0x000000ff,
                                                 0x0000ff00,
                                                 0x00ff0000,
-                                                0xff000000);
+                                                0xff000000));
 #endif
```
v1993, line 95: Big endian path apparently not fully updated? I think this now should become

```c++
#if SDL_BYTEORDER == SDL_BIG_ENDIAN
SDL_PixelFormat format = ...
#else
SDL_PixelFormat format = ...
#endif
SDL_Surface* surface = SDL_CreateSurface((int) width, (int) height, format);
```

Although idk if the big endian path is actually tested on any machine out there.

src/lincity-ng/EconomyGraph.cpp
```diff
@@ -188,9 +188,9 @@ EconomyGraph::drawHistoryLineGraph(Painter& painter, Rect2D space) {
       space.p2 - Vector2(i, 0)));
 
     // glEnable(GL_LINE_SMOOTH);
-    val = std::max(0.f, log((float)pop) + popScale0) * popScale1;
+    val = std::max(0.0, log((float)pop) + popScale0) * popScale1;
     valP = !popPrev ? val :
-      std::max(0.f, log((float)popPrev) + popScale0) * popScale1;
+      std::max(0.0, log((float)popPrev) + popScale0) * popScale1;
```
v1993: Seems unrelated? You've also changed the literal type from float to double, which seems undesirable.

src/lincity-ng/main.cpp
```diff
@@ -178,7 +175,7 @@ void initVideo(int width, int height)
     else
 #endif
     {
-        window_renderer = SDL_CreateRenderer(window, -1, 0);
+        window_renderer = SDL_CreateRenderer(window, NULL);
```
v1993: `nullptr` is preferred in C++.

src/lincity-ng/Sound.cpp
```diff
@@ -248,15 +248,24 @@ Sound::Sound()
...
+    if (NULL == mixer) {
         fprintf(stderr, "Couldn't open audio: %s\n", SDL_GetError());
         return;
     } else {
         audioOpen = true;
         loaderThread = SDL_CreateThread(soundThread, "Sound", this);
     }
 
+    musicTrack=MIX_CreateTrack(mixer);
+    soundTrack=MIX_CreateTrack(mixer);
```
v1993:
```suggestion
    musicTrack = MIX_CreateTrack(mixer);
    soundTrack = MIX_CreateTrack(mixer);
```

src/lincity-ng/Sound.cpp
```diff
@@ -437,15 +448,17 @@ Sound::playMusic()
...
             std::cerr << "Couldn't load music file '"
                 << currentTrack.filename << "': "
                 << SDL_GetError() << std::endl;
             return;
         }
 
-        Mix_PlayMusic(currentMusic, 1);
+        MIX_SetTrackAudio(musicTrack, currentMusic);
+        MIX_PlayTrack(musicTrack, 0);
+//        Mix_PlayMusic(currentMusic, 1);
```
v1993:
```suggestion
```

Commented out code.

mk/iwyu.map
```diff
@@ -4,89 +4,89 @@
...
-{ include: [ <SDL_touch.h>, private, <SDL.h>, public ] },
-# { include: [ <SDL_ttf.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_types.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_version.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_video.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_vulkan.h>, private, <SDL.h>, public ] },
+# { include: [ <SDL3/SDL2_framerate.h>, private, <SDL3/SDL.h>, public ] },
+# { include: [ <SDL2_gfxPrimitives.h>, private, <SDL3/SDL.h>, public ] },
+# { include: [ <SDL2_imageFilter.h>, private, <SDL3/SDL.h>, public ] },
+# { include: [ <SDL3/SDL2_rotozoom.h>, private, <SDL3/SDL.h>, public ] },
```
v1993: Still referencing SDL2. Probably go ahead and remove these commented out lines.

src/gui/Event.hpp
```diff
@@ -19,7 +19,7 @@ Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
 #ifndef __EVENT_HPP__
 #define __EVENT_HPP__
 
-#include <SDL.h>        // for SDL_Event, SDL_Keysym
+#include <SDL3/SDL.h>   // for SDL_Event, SDL_Keysym
```
v1993: `SDL_Keysym` is gone in SDL3.

v1993 (changes requested):
Thank you for your work! I've noted some things that could use a little further refinement from a quick review. Haven't tested yet.

mk/cmake/modules/SDL2/FindSDL2.cmake
myoung008: These finders aren't needed on my distro (Gentoo), and it looks like it may be that way for others with SDL3 packaging.  Removing the files.

src/gui/PainterSDL/TextureManagerSDL.cpp
myoung008: Yes, though I've fixed the comment so that it is correct now

src/lincity-ng/main.cpp
myoung008: haha! Yep.  I do a lot in regular c too.

src/gui/Event.cpp
myoung008: removed version check and alternate code path

src/gui/Event.hpp
myoung008: fixed comment.  SDL_Keysym was already removed from code.

src/lincity-ng/EconomyGraph.cpp
myoung008: This was intentional, although unrelated to SDL3.  My compiler would not compile as it was originally.  no version of std::max for float, double

myoung008:
Unless I've missed something I've applied all of your suggestions.  I also fixed CreateRGBSurface in the GL path, but it appears that is incomplete and old anyway.

src/lincity-ng/EconomyGraph.cpp
myoung008: I checked the docs and <cmath> log returns a double.  As of GCC15.2.1 the 0.f constant does not compile because of the datatype difference sent to std::max.  I've left it as a double for now.  If you want me to change it back I can.  It seems that this change will be coming whether a part of the SDL3 patch or not.

src/lincity-ng/main.cpp
dbear496: The [documentation](https://wiki.libsdl.org/SDL3/SDL_CreateRenderer) for `SDL_CreateRenderer` says to use `NULL`, and for this reason alone, I have preferred using `NULL` in several places throughout the code. However, it's not a deal-breaker either way.

src/lincity-ng/EconomyGraph.cpp
dbear496: Which docs did you check? cppreference says `log(float)` returns float. Everything in LinCity-NG uses float, so for now I think it's best to keep it that way as much as possible.

src/lincity-ng/main.cpp
dbear496: Looks like the return type of `log` is implementation-defined for C++23 and upward, but this shouldn't be an issue yet because Lincity-NG [explicitly uses C++17](https://github.com/lincity-ng/lincity-ng/blob/46d50dbd62d36568f902e699e207c05f824c78e7/CMakeLists.txt#L102)...unless your compiler decided to use C++23 anyway. When we implement support for C++23, we should probably switch to use `logf` wherever `log` is currently used.

mk/cmake/modules/SDL2/FindSDL2.cmake
dbear496: My CMake installation includes these find modules and they are SDL3 compatable. I'm not sure when they were added as part of CMake. Potentially people with an old version of CMake may not have the needed find modules and run into issues, but in any case, find modules for SDL2 aren't going to help with finding SDL3, so deleting these is a good call.

src/lincity-ng/EconomyGraph.cpp
dbear496: Looks like the return type of `log` is implementation-defined for C++23 and upward, but this shouldn't be an issue yet because Lincity-NG [explicitly uses C++17](https://github.com/lincity-ng/lincity-ng/blob/46d50dbd62d36568f902e699e207c05f824c78e7/CMakeLists.txt#L102)...unless your compiler decided to use C++23 anyway. When we implement support for C++23, we should probably switch to use `logf` wherever `log` is currently used.

src/gui/PainterSDL/PainterSDL.cpp
```diff
@@ -98,8 +98,9 @@ void
 PainterSDL::fillPolygon(int numberPoints, const Vector2* points) {
   assert(numberPoints >= 3);
 
-  SDL_Color color = (SDL_Color)
-    {.r = fillColor.r, .g = fillColor.g, .b = fillColor.b, .a = fillColor.a};
+  SDL_FColor color = (SDL_FColor)
+    {.r = 1.0f * fillColor.r / 255, .g = 1.0f * fillColor.g / 255,
+     .b = 1.0f * fillColor.b / 255, .a = 1.0f * fillColor.a / 255};
```
dbear496:
```suggestion
    {.r = fillColor.r / 255.f, .g = fillColor.g / 255.f,
     .b = fillColor.b / 255.f, .a = fillColor.a / 255.f};
```

src/gui/Gradient.cpp
dbear496: Hmm. Is there some reason not to just use `SDL_PIXELFORMAT_RGBA8888`/`SDL_PIXELFORMAT_ABGR8888`?

src/gui/Paragraph.cpp
```diff
@@ -347,9 +347,9 @@ Paragraph::resize(float width, float height)
             if(spanimages.size() == 1) {
                 lineimages.push_back(spanimages.back());
             } else {
-                SDL_Surface* lineimage = SDL_CreateRGBSurface(0, (int) pos.x,
-                        (int) lineheight, 32,
-                        0x000000ff, 0x0000ff00, 0x00ff0000, 0xff000000);
+                SDL_Surface* lineimage = SDL_CreateSurface((int) pos.x,
+                        (int) lineheight, SDL_GetPixelFormatForMasks(32,
+                        0x000000ff, 0x0000ff00, 0x00ff0000, 0xff000000));
```
dbear496: Can `SDL_PIXELFORMAT_ABGR8888` be used instead of `SDL_GetPixelFormatForMasks`?

src/gui/Paragraph.cpp
```diff
@@ -420,16 +420,16 @@ Paragraph::resize(float width, float height)
...
     }
 
     /* Step2: compose all lines to the final image */
     if(width < 0) {
         width = lineimages[0]->w;
     }
-    SDL_Surface* result = SDL_CreateRGBSurface(0, (int) width, (int) height,
-            32, 0x000000ff, 0x0000ff00, 0x00ff0000, 0xff000000);
+    SDL_Surface* result = SDL_CreateSurface((int) width, (int) height,
+            SDL_GetPixelFormatForMasks(32, 0x000000ff, 0x0000ff00, 0x00ff0000, 0xff000000));
```
dbear496: `SDL_PIXELFORMAT_ABGR8888`

src/gui/Paragraph.cpp
```diff
@@ -446,15 +446,15 @@ Paragraph::resize(float width, float height)
         }
         rect.y = (Sint16) ycoords[i];
         SDL_BlitSurface(lineimages[i], 0, result, &rect);
-        SDL_FreeSurface(lineimages[i]);
+        SDL_DestroySurface(lineimages[i]);
     }
-    SDL_Surface* surface = SDL_ConvertSurfaceFormat(result, SDL_PIXELFORMAT_RGBA8888, 0);
-    SDL_FreeSurface(result);
+    SDL_Surface* surface = SDL_ConvertSurface(result, SDL_PIXELFORMAT_RGBA8888);
```
dbear496, line 451: Unrelated to SDL3 migration: why not create the surface with the correct pixel format to begin with?

src/lincity-ng/Sound.cpp
```diff
@@ -248,15 +248,24 @@ Sound::Sound()
...
     audioOpen = false;
+    if (!MIX_Init()) {
+      SDL_Log("MIX_Init failed: %s", SDL_GetError());
+    } else {
+      SDL_Log("SDL_mixer is ready!");
+    }
     /* Open the audio device */
-    if (Mix_OpenAudio( MIX_DEFAULT_FREQUENCY, MIX_DEFAULT_FORMAT, 2, 2048) < 0) {
+    mixer = MIX_CreateMixerDevice(SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK, NULL);
+    if (NULL == mixer) {
```
dbear496:
```suggestion
    if(!mixer) {
```

src/lincity-ng/Sound.hpp
```diff
@@ -73,20 +73,25 @@ class Sound
 
     void setTechLevel(int tech) { tech_level = tech; }
 
+    MIX_Track *musicTrack;
```
dbear496: Is there a reason this is public while `soundTrack` is private?

dbear496 (commented):
I haven't tested it yet, but here are some comments I had on my first look through.

src/lincity-ng/Sound.hpp
myoung008: Yes, because of main.cpp:297 MIX_SetTrackStoppedCallback(sound->musicTrack...
The SDL2 code effectively used a singleton that is not available in SDL3.

dbear496: Can we move `musicHalted` to Sound.cpp and move the call to `MIX_SetTrackStoppedCallback` to `Sound::Sound`?

mk/iwyu.map
```diff
@@ -4,89 +4,85 @@
...
-{ include: [ <SDL_test_random.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_thread.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_timer.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_touch.h>, private, <SDL.h>, public ] },
-# { include: [ <SDL_ttf.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_types.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_version.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_video.h>, private, <SDL.h>, public ] },
-{ include: [ <SDL_vulkan.h>, private, <SDL.h>, public ] },
+{ include: [ <SDL3/SDL_assert.h>, private, <SDL3/SDL.h>, public ] },
```
dbear496, line 7: Add `SDL3/SDL_init.h` and `SDL3/SDL_properties.h` here.

dbear496 (changes requested):
Again, thanks for the contribution. I briefly tested it, and it seems to work well. I just noticed 2 new warnings from IWYU that should be addressed in addition to my code-review comments from Saturday.

myoung008:
Current set of changes include all suggestions. 
I went ahead and changed the doubles back to floats, using logf.  There were only 4 instances of log() and all in the same file. logf is available since C++11.
musicHalted moved to Sound as a static member.  It must be static or classless for the SDL3 callback.

dbear496:
@myoung008 I think you misunderstood me; I meant to update the IWYU mappings with SDL_init.h and SDL_properties.h. I pushed a commit to address it.

dbear496:
Looks good to me. Only thing remaining is to update [the changelog](https://github.com/lincity-ng/lincity-ng/blob/master/CONTRIBUTING.md#changelog), [creditslist.xml](https://github.com/lincity-ng/lincity-ng/blob/master/data/gui/creditslist.xml), and [copyright statements](https://github.com/lincity-ng/lincity-ng/blob/master/CONTRIBUTING.md#attribution-copyright-and-licensing).

myoung008:
I think it's done.  I didn't add myself to copyright on every file I touched.  Many were just updating headers.

src/gui/PainterSDL/PainterSDL.cpp
```diff
@@ -22,7 +23,8 @@
 
 #include "PainterSDL.hpp"
 
-#include <SDL.h>           // for SDL_GetError, SDL_SetRenderDrawColor, SDL_...
+#include <SDL3/SDL.h>      // for SDL_GetError, SDL_SetRenderDrawColor, SDL_...
+#include <SDL3/SDL_properties.h> // for SDL_PropertiesID, SDL_GetNumberPrope...
```
dbear496:
```suggestion
```

This header is provided by SDL3/SDL.h.

src/lincity-ng/main.cpp
```diff
@@ -22,9 +23,10 @@
 
 #include "main.hpp"
 
-#include <SDL.h>                                 // for SDL_GetError, SDL_Se...
-#include <SDL_mixer.h>                           // for Mix_HookMusicFinished
-#include <SDL_ttf.h>                             // for TTF_Init, TTF_Quit
+#include <SDL3/SDL.h>                            // for SDL_GetError, SDL_Se...
+#include <SDL3/SDL_init.h>                       // for SDL_INIT_AUDIO, SDL_...
```
dbear496:
```suggestion
```

provided by SDL3/SDL.h

src/lincity-ng/main.cpp
```diff
@@ -22,9 +23,10 @@
...
+#include <SDL3_mixer/SDL_mixer.h>                // for Mix_HookMusicFinished
```
dbear496:
```suggestion
```

unused

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Combined the source and GUI translation PO files into one.
 - Migrated GUI module from libxml2 to libxml++.
 - Added locale-independent formatting and parsing for XML.
+- Migrated from SDL2 to SDL3; Requires SDL >= 3.20.0
```
dbear496: Please move this to a new file in CHANGELOG.d; see [CONTRIBUTING.md](https://github.com/lincity-ng/lincity-ng/blob/master/CONTRIBUTING.md#changelog).

dbear496: Did you mean "3.2.0"?

dbear496 (changes requested):
Just a few simple issues. Also the SDL version numbers in doc/DEPENDENCIES.md need to be updated.

myoung008:
I think I've covered everything you've mentioned.

dbear496 (dismissed):

CHANGELOG.d/sdl3.md
```diff
@@ -0,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+
+###### Internal
+- Migrated from SDL2 to SDL3; Requires SDL >= 3.20.0
```
v1993:
```suggestion
- Migrated from SDL2 to SDL3; Requires SDL >= 3.2.0
```

doc/DEPENDENCIES.md
```diff
@@ -4,19 +4,19 @@ Following is the list of dependencies for building and/or running LinCity-NG.
...
 
-* SDL2_image 2.0.0 or later (with png support)
+* SDL3_image 3.0.0 or later (with png support)
 
   http://www.libsdl.org/projects/SDL_image/
 
-* SDL2_ttf 2.0.12 or later
+* SDL3_ttf 3.0.0 or later
 
   http://www.libsdl.org/projects/SDL_ttf/
```
v1993:
```suggestion
* SDL3_mixer 3.2.0 or later (with Ogg Vorbis support)

  https://github.com/libsdl-org/SDL_mixer

* SDL3_image 3.2.0 or later (with PNG support)

  https://github.com/libsdl-org/SDL_image

* SDL3_ttf 3.2.0 or later

  https://github.com/libsdl-org/SDL_ttf
```

3.2.0 is the first stable release for each of these libraries; update URLs to their redirect targets while we're at it.

v1993 (changes requested):
Code LGTM; slight tweaks to docs and this should be good to go.

doc/DEPENDENCIES.md
myoung008: I made the requested changes.

v1993 (approved):
Thank you!

myoung008:
That last push had no changes, but commits are signed now.

v1993:
@dbear496 FYI: on Flatpak we now need to build SDL3_mixer and SDL3 itself (since version in the runtime is too old for SDL3_mixer); see https://github.com/flathub/io.github.lincity_ng.lincity-ng/tree/pure-sdl3. This won't be needed once runtime 26.08 releases later this year.

dbear496:
> FYI: on Flatpak we now need to build SDL3_mixer and SDL3 itself (since version in the runtime is too old for SDL3_mixer); see https://github.com/flathub/io.github.lincity_ng.lincity-ng/tree/pure-sdl3. This won't be needed once runtime 26.08 releases later this year.

@v1993, I assume that doesn't require any changes to this repo -- but rather the Flathub release repo. You might also have to build SDL3_ttf so it links with the built version of SDL3?

v1993:
Yep, I've already implemented the required changes in the linked branch. SDL TTF and Image libraries don't seem to need a rebuild since SDL maintains ABI compatibility and links again non-version specific symlink (`libSDL3.so.0` rather than e.g. `libSDL3.so.0.4.8`).

PR #349, opened 2026-05-09
Fix a duplicate key in desktop file

v1993:
Fixup for #341. Please be a bit more careful with the desktop file, Flatpak build actually validates those - I'll include the patch downstream for now, but idk how desktop environments will handle this in other builds.


dbear496 (approved):

dbear496:
My bad. I should have caught this when reviewing #341.

PR #352, opened 2026-05-11
fix map shimmering

### LINKED ISSUE #332: Use mipmaps for better zooming out
Due to the lack of mip-mapping you can observe grainy/aliased look while significantly zoomed out and shimmering when moving around in that state.

Unfortunately, neither SDL2 nor SDL3 currently provide a good way to avoid that; there was a [shot](https://github.com/libsdl-org/SDL/pull/10254) at adding mipmapping functionality, but so far it hasn't progressed beyond the initial attempt. This means that there are only so many options for LinCity-NG:

* Wait for SDL3 to add support for mipmaps (and SDL_mixer/SDL_image get an SDL3 version) - can take a while, seemingly little progress on that, but would be the ideal solution
* Use `SDL_GL_BindTexture`+`glGenerateMipmap` like suggested on https://github.com/libsdl-org/SDL/issues/4156; this is the simplest solution, but won't work with non-OpenGL renderers (e.g. direct3d)
* Switch back to the direct OpenGL renderer - I take that's undesirable
* Switch to SDL3 and use SDL_GPU to implement a 2D renderer (and SDL_mixer/SDL_image get an SDL3 version) - potentially limits target hardware due to needing Vulkan, more work than other options, still requires waiting
* Implement mipmaps manually, i.e. generate multiple textures and select the most appropriate one depending on resolution - probably less efficient and requires extra work, but could be implemented without other compromises

dbear496:
Fixes map shimmering when panning by rounding the viewport position.

resolves #332

### DIFF
--- a/src/gui/Vector2.hpp
+++ b/src/gui/Vector2.hpp
@@ -23,6 +23,7 @@
 #ifndef __VECTOR_HPP__
 #define __VECTOR_HPP__
 
+#include <cmath>
 #include <ostream>
 
 class Rect2D;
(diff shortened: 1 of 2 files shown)


dbear496:
This "fix" doesn't properly handle high-dpi display (#357).

dbear496:
Superseded by #357

PR #353, opened 2026-05-12
Update french translation

Dadu042:
Last updated in 2008 or 2011.

I am not a .PO editing expert, please recheck it.

### DIFF
--- a/CHANGELOG.d/french.md
+++ b/CHANGELOG.d/french.md
@@ -1,0 +1,10 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+
+###### Internal
+
+###### Documentation / Translation
+- Updated French translation.
--- a/data/gui/creditslist.xml
+++ b/data/gui/creditslist.xml
@@ -183,6 +183,10 @@
     <span style="year">2005</span>
   </p>
   <p style="person">fr –
+    Dadu042
+    <span style="year">2026</span>
+  </p>
+  <p style="person">fr –
     Laurent Rocher
     <span style="year">2006</span>
   </p>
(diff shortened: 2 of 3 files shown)


dbear496:
I noticed that the PO file used absolute file paths in the comments, so I went ahead and converted those to relative. The absolute paths were caused by a bug in TranslateTarget.cmake, which I fixed just now but haven't yet merged to master.

Just doing a quick peruse through the PO file; it looks like there are spaces inserted before all the punctuation. Is that standard for French or is it an artifact from your PO editor? These spaces could cause the paragraph layout code to insert a line break immediately before punctuation.

Dadu042:
I do not see what spaces you are talking about, please give me 2 or 3 examples.

dbear496:
To give a few examples, see lines [441](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L441), [453](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L453), [466](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L466), [476](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L476), [490](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L490), [549](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L549), [805](https://github.com/Dadu042/lincity-ng/blob/9bf32f685253f595d668e48814950d930e9451b6/data/locale/fr.po#L805).

Dadu042:
Thanks. All these spaces are normal in french, this is how we write.

PR #357, opened 2026-05-19
support high-dpi/scaled display

### LINKED ISSUE #319: Windows version does not honor window size/mode from config file
In the Windows version of LinCity-NG running under Wine, after setting the window size to non-default and restarting the game, LinCity-NG still uses the default window size. If fullscreen is toggled on, game is restarted, fullscreen is toggled off, and game restarted again, then LinCity-NG gets stuck in fullscreen mode.

### LINKED ISSUE #333: Graphics are too small
Hello!
I use LinCity (2.14.2) on Endeavouros (current) and a 4K monitor.
Unfortunately, I have found that the graphics are so small that they are barely visible.

It would be great if there was a setting option, but unfortunately I can't find anything like that.
Could you adjust the graphics so that they are easy to see on a 4K monitor?

Best regards, R. Lehmeier

dbear496:
Adds support for high-dpi/scaled displays by divorcing the window and painting coordinate system from the pixel size.

Also adds support for slanted gradients and slants the gradient backgrounds for window titles.

resolves #333


paulorlyk:
Main menu is zoomed in and cropped on X11

```
$ xrdb -query | grep dpi
Xft.dpi:	192
```

dbear496:
I realized that I didn't enable the `SDL_WINDOW_HIGH_PIXEL_DENSITY` flag, so I'm fixing that now.

src/lincity-ng/MapPoint.hpp
paulorlyk, line 1: Should probably also be removed from src/lincity-ng/CMakeLists.txt

paulorlyk:
Unfortunately it is still the same



(pw.conf stuff is unrelated)

dbear496:
@paulorlyk Is it any better now on X11? It looks good to me on Wayland.

paulorlyk:
Unfortunately there is no changes for me. Still exactly the same.

dbear496:
@paulorlyk , how is it now?

paulorlyk:
Main menu look fine, but buttons respond to mouse moves and clicks only in marked area (approximately):


Also it crashes when attempting to start the game:
```
$ cmake -B build -DCMAKE_BUILD_TYPE=Release
$ cmake --build build --parallel
$ ./build/bin/lincity-ng --app-data build/share/lincity-ng
...
terminate called after throwing an instance of 'std::runtime_error'
  what():  couldn't load image: Couldn't open build/share/lincity-ng/images/gui/buttonpanel/menu/game_menu.png: No such file or directory
```
Not sure if I am doing it wrong...

dbear496:
Alright. I applied some scaling to mouse events, so hopefully this should be all good on X11. It would be great to have someone test this.

paulorlyk (changes requested):
The good news is that it seems like scaling works. The bad news, and I really hate to say this, but it seems like it scales in the wrong direction.

It's now squeezed into 1/16 of the window, not 1/4 as it was before. The entire window could be covered by mouse while staying in the approximate area marked red on the picture below.

Image

dbear496:
Oops! I reversed the scaling directions. Hopefully it works now.

paulorlyk (approved):
Works! Seems to be playable in a quick 4 minute test!

The only thing is that upon returning to the main menu from the game, the scaling is wrong, looks exactly like on a picture from my very first message. Attempt to resize a window immediately fixes the issue.

Also it seems like window can be resized to 1x1 size, after which game will not start again:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  error: value 1 is outside range 640..2147483647
```
Editing lincity-ng.conf helps.

paulorlyk:
FYI
After removing `<x>` and `<y>` from lincity-ng.conf, game looks like this:



But it's recoverable from there, just default values might need scaling too.

dbear496:
I think I fixed the scaling when returning to the main menu. The crash after sizing the window too small should be addressed in #376 . The window being the wrong default size is an issue I need to address.

paulorlyk:
Can confirm, now the scaling is fine upon returning to the main menu.

dbear496:
I fixed a couple issues relating to the window size config. Now on X11 the default window size should adjust according to the display content scale. And on Wayland the window will stay a consistent size between restarts. I also refactored some stuff, so hopefully I didn't break anything.

Justus-Fischer:
> The crash after sizing the window too small should be addressed in #376 .

Yes, I agree. I will fix this in my local copy and include it in my current PR (#376).

One other thing I noticed in the error message:

>  error: value 1 is outside range 640..2147483647

Currently, the minimum value for `videoX` is set to 640 and for `videoY` to 480.
However, the smallest selectable size in the menu is actually 320x200.

My thought was to change the minimums to 320 and 200 to match the UI, or did you have different minimum values in mind?

dbear496:
> Currently, the minimum value for videoX is set to 640 and for videoY to 480.
> However, the smallest selectable size in the menu is actually 320x200.

The sizes available in the menu are platform dependent (See `MainMenu::changeResolution`). When we query the display for available resolutions, we should filter out the ones that are too small.

Another gotcha to be aware of: the resolution in the config file is in _window coordinates_, so scaled displays on Windows, X11, and Android require a larger (or smaller, if the scale is <100%) window from the perspective of the config file.

paulorlyk (approved):
Default window size works.

Minor stuff:

New game menu on default window size:

Image

Independent of window size - font is clipped:

Image

Right bottom corner - font is very small, barely readable:

Image

dbear496:
I fixed the label size on the economy graph. The other issues are unrelated to display scaling.

PR #376, opened 2026-08-21
Improve input validation to fix #366

### LINKED ISSUE #366: Can pick 0 or negative map size in settings
In the game settings, you can reduce the map size as much as you want, even picking 0 or a negative map size. If you do so and try to start a game, it crashes. Even worse, it will insta-crash in subsequent starts, locking you out of the game. The user is then forced to manually fix the config file (which most players won’t figure out).

Discovered in: 431c9639df39c62e3ec159f18e8f3e1b0b2c6a65

Justus-Fischer:
@Wuzzy2 found a bug where the user can change the map size to 0 or a negative size, which caused the game to crash.

While `src/lincity-ng/Config.cpp` (line 223) attempts to protect the game from negative values, it does not fully prevent invalid inputs via the UI. Therefore, `std::clamp` has been implemented in `src/lincity-ng/MainMenu.cpp` (line 571), which handles the logic for the buttons that set the map size. This now protects the game from crashes by using the min and max limits from `Config.cpp` (line 224).
To simplify the code, the constants `worldSizeMin` and `worldSizeMax` have been added to `src/lincity-ng/Config.hpp` (lines 76 and 77) to replace the values that were previously written directly into `Config.cpp`.

Now it is impossible for the user to select a map size lower than 50.

Fixes #366

### DIFF
--- a/CHANGELOG.d/fix-negative-map-size.md
+++ b/CHANGELOG.d/fix-negative-map-size.md
@@ -1,0 +1,11 @@
+## Unreleased
+
+###### Gameplay
+
+###### User Interface
+- Fixed a bug where users could pick a negative map size in settings
+
+###### Internal
+- Added the constants worldSizeMin and worldSizeMax in Config.hpp 
+
+###### Documentation / Translation
--- a/data/gui/creditslist.xml
+++ b/data/gui/creditslist.xml
@@ -42,6 +42,10 @@
     <span style="year">2005 - 2007</span>
   </p>
   <p style="person">
+    Justus Fischer
+    <span style="year">2026</span>
+  </p>
+  <p style="person">
     Edu Garcia
     <span style="year">2023</span>
   </p>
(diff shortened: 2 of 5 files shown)


Justus-Fischer:
Sorry for the bot review, I didn't even know that this setting existed but now I have deactivated it.

dbear496 (changes requested):
This is not my favorite solution to the issue. Input validation is something that needs to be done for most config options, so it should be done in a way that is uniform across all options. And adding `worldSizeMin`/`worldSizeMax` variables to the `Config` class is not a uniform practice across options. For example, music and FX volume also needs to be validated in the range 0-100.

I would prefer input validation be part of the `Config::Option` class with a `set` method that verifies the input before changing the value and throws a custom subclass of `std::runtime_error` for validation errors. Perhaps the most straight-forward way to do this is for the `Option` constructor to accept a `std::function<bool, const T&>` optional argument. Alternatively, `Option<int>` could be specialized to have min/max allowable values, which could be used for disabling the spin buttons in the UI when the min/max is hit. In addition to a `set` method, `trySet`, `setConfig`, and `trySetConfig` may be useful.

Justus-Fischer:
Thanks for the feedback. I will change it accordingly.

PR #383, opened 2026-08-29
[WIP] Update Dutch translation

### LINKED ISSUE #380: improving the Dutch translation
I would like to improve the Dutch translation for the project.
And this issue is to let other people know that I'm working on it.

subbink:
This is still a Work in Progress to update the Dutch translation.
There is still more work to be done before it is finished.


v1993:
A lot of changes replace hard tabs with spaces. You might want to change your editor settings to avoid this.

subbink:
I have changed a lot of files to be more like the english .xml help files.
for example the [data/help/nl/export.xml](https://github.com/lincity-ng/lincity-ng/blob/44fcdb0adc02fd961fed1a0d286ab7892df085b7/data/help/nl/export.xml#L10) has a change from a tab to 2 spaces, because [the original english file](https://github.com/lincity-ng/lincity-ng/blob/12c5b557229e3b644cd1687651232e992b6b4b20/data/help/en/export.xml#L10) has 2 spaces at the same place.

This makes it easier for me to compare the files, but if those changes are not wanted, I can revert it back to the way they were.

v1993:
Considering English files are automatically generated from a template this should be fine (although I must question the generator). It'd probably be optimal to have indentation changes as a separate commit, but I believe it's not a must.

v1993:
Heads-up: you forgot to update the language code for comment in desktop file.

subbink:
> Heads-up: you forgot to update the language code for comment in desktop file.

Thank you for mentioning, I have fixed this in my local copy, so it will be part of my next commit.

subbink:
While doing the translation, I saw a situation which is a bit hard to translate.
On [L1585 of GameView.cpp](https://github.com/lincity-ng/lincity-ng/blob/99939d94483189b5d512732b27555cd3e2e937a1/src/lincity-ng/GameView.cpp#L1585) there is an original text " To build `number` of them will cost about `number`$" but this is not one string to be translated, but 4 different ones.
If I would translate the full sentence to Dutch it would be something like: " Om `number` ervan te bouwen zal ongeveer €`number` kosten". But this seems very strange if you look at the translated strings.
The different strings would be translated as follows:
" To build" -> " Om"
" of them " -> " ervan te bouwen "
"will cost about " -> "zal ongeveer €"
"$" -> "kosten"

This is for example caused by the placement of the currency sign at a different location of the number between the languages, but also the verbs are in different places in the sentence.

In other strings (for example [Game.cpp#L965](https://github.com/lincity-ng/lincity-ng/blob/99939d94483189b5d512732b27555cd3e2e937a1/src/lincity-ng/Game.cpp#L965)) I see the use of `{}` which would make it easier to translate a complete sentence.

What would be the best way to improve this situation?

dbear496:
> While doing the translation, I saw a situation which is a bit hard to translate.

The gettext manual advises against breaking up strings like that for exactly this reason. I've been slowly updating such occurrences to use printf-style strings when I touch related lines of code, but this specific one hasn't been updated yet. Since it's causing you issues, I'll go ahead and update this one.

dbear496:
Actually, this string is never presented to the user, so there's no use translating it anyway. The string on line 1595, however, is shown to the user, and it has the same issue. *sigh* This whole section of code really needs to be refactored--GameView isn't even the right place for it.

dbear496:
@subbink , I updated the messages that you were having trouble with in #384 , so you should be able to translate them more easily.

subbink:
thank you for the changes, I will use the new strings in my work.

PR #387, opened 2026-09-14
Fix lowerLeft terrain-height adjustment

FabioPedretti:
The lowerLeft declared here shadowed the outer one used by getTile()
right after, so the result was discarded. The formula was also
wrong, it used getHeight() instead of map.alt_max.


dbear496:
Why is it squaring the height? I would expect something more like this:
```
lowerLeft.y += getWorld().map(mp)->alt_max * scale3d * zoom / getWorld().map.alt_step;
```

FabioPedretti:
FTR, this was introduced in 6c8616a69.
Make sense, I updated the commit with your proposal, thanks.

dbear496:
The issue with tiles close to the bottom of the screen disappearing is fixed now. But when I was testing, I noticed that I cannot pan to the top of the map when zoomed in.

dbear496 (approved):

Issue #226, opened 2024-12-03
macOS builds


trufae:

thomashrabe:
Hi there, I got it to build on Mac, running it segfaults for now though...

dbear496:
To diagnose issues at runtime (e.g. segfaults), I recommend using build type `Debug`, `DebugOpt`, `RelWithDebuInfo`, or `BetaTest` (as you have done) and running the game in a debugger such as gdb. All these build types enable debug symbols in the executable, so running in the debugger is nicer. The build type is case insensitive, so specifying `debug` instead of `Debug` as you have done is fine. Running in gdb is not difficult: 1) from the command line, run `gdb build/bin/lincity-ng`, 2) at the gdb prompt, enter `r` to start the game, 3) when the program crashes enter `bt` to show the call backtrace, and 4) to exit gdb, enter `q`. Of course there is so much more you can do with gdb, but those are the essentials. Once you have the backtrace, you can post it here for me to take a look at.

thomashrabe:
Thanks for the build flags, thats what I didnt find in the docs. Yes, I have experience with `gdb` / `ddd` so that should work. Maybe I'll get the debugger running in VSCode, lets see. Thank you!

thomashrabe:
@dbear496 I see with your linked PR that this may be actually obsolete?

thomashrabe:
@dbear496 
It is failing here
```
if(!foundRodd) {
      std::ostringstream msg;
      msg << "Failed to mount read-only data directory '"
          << appdatadir << "' or '"
          << INSTALL_FULL_APPDATADIR
          << "': " << getPhysfsLastError();
      throw std::runtime_error(msg.str());
    }
```

I am not sure what the game needs, or PhysFS. I changed the folder I am starting the game to `777` yet it still segfaults because `appdatadir` is undefined.

dbear496:
@thomashrabe See if PR #234 helps at all. You can get those changes via the following commands:
```
git remote add dbear496 https://github.com/dbear496/lincity-ng.git
git checkout --track dbear496/extricate-physfs
```

This PR refactors the directory discovery code, and it also allows configuring the app data directory manually.

dbear496:
According to the comment thread in #243, LinCity-NG version 2.14.0 should work on MacOS  as long as the system has proper CMake config modules.

PR #388, opened 2026-09-15
Fix build on macOS

kelvSYC:
## Summary

This change contains two small changes that are needed to build LinCity-NG on macOS, without affecting the Linux compilation path. This change relates to issue #226, and contains the following:
* `std::hash` resolves ambiguously on macOS and fails to build. This is replaced with something that is portable. IWYU updated to account for the difference between libstdc++ (used on Linux) and libc++ (used on macOS).
* one where CoreServices was not properly linked, which is needed for the proper functionality of `cfgpath.h` on macOS.

Additionally, `pkg-config` is explicitly marked as a build dependency.

Because the CoreServices APIs are deprecated, and `cfgpath.h` is vendored, no attempt is made to fix or work around the deprecation issue or the fact that it causes configuration files be placed contrary to Apple's best practices.

The process to build a redistributable macOS application bundle is deliberately omitted, due to the additional codesigning, notarization, and other requirements. These details are deliberately left for future work, but should be enough to, say, configure GitHub Actions to build the application bundle with a macOS runner if and when these details are sorted out.

## Test plan

- [x] `cmake -B build -DCMAKE_BUILD_TYPE=Release` configures successfully on macOS
- [x] `cmake --build build --parallel` builds successfully to completion
- [x] Resulting binary links and runs (`--version`)
- [x] Repeated the above in a minimal environment (`env -i` with a bare `PATH`, no shell profile) to approximate a non-interactive CI shell

Generative AI (Claude, Sonnet 5, via Claude Code) was used to diagnose the build failures and implement, document, and verify the fix; this description was written by the human contributor.


doc/BUILD_OPTIONS.md
```diff
@@ -84,6 +81,21 @@ inadvertently break it by moving/copying the binary away from other installed
...
+
+Code signing and notarization specifically are out of scope: they only
+matter for binaries distributed to other users (Gatekeeper checks binaries
+downloaded via a browser or similar), not for binaries built locally from
+source. One relevant note for that future work: save data and settings are
+already written to `~/Library/Application Support` (see `cfgpath`), not
+alongside the installed binary, which avoids conflicts with a signed
+bundle's contents.
+
+
```
dbear496: I don't think this is helpful since it only explains what is *not* supported. At the very least, it doesn't belong in this file, which is specifically for documenting CMake variables.

lib/cfgpath/CMakeLists.txt
```diff
@@ -7,3 +7,10 @@ set_target_properties(cfgpath PROPERTIES
   LINKER_LANGUAGE C
   C_INCLUDE_WHAT_YOU_USE ""
 )
+
+# macOS: cfgpath.h's __APPLE__ branch uses Carbon FSFindFolder/FSRefMakePath
+# (CoreServices) to locate ~/Library/Application Support. Link the framework so
+# consumers resolve those symbols.
```
dbear496: This is comment wordy, and it can be removed entirely.

src/lincity/MapPoint.hpp
```diff
@@ -27,9 +27,7 @@
 #include <cstddef>        // for size_t
 #include <ostream>        // for ostream
 
-namespace std {
-template<class Key> struct hash; // IWYU pragma: keep
-}  // namespace std
+#include <functional>  // for std::hash (libc++ inline-namespace safe; upstream forward-declares std::hash which is ambiguous under libc++'s std::__1)
```
dbear496: IWYU prefers `<string_view>`.

doc/BUILD_OPTIONS.md
dbear496: I don't think this is helpful since it only explains what is *not* supported. At the very least, it doesn't belong in this file, which is specifically for documenting CMake variables.

doc/DEPENDENCIES.md
```diff
@@ -79,6 +79,29 @@ Following is the list of dependencies for building and/or running LinCity-NG.
...
+series. `brew install libxml++` alone will not satisfy this project's
+`libxml++-5.0` requirement.
+
+The macOS SDK also ships its own old copy of libxml2 (2.9.13), which CMake may
+find in preference to the newer Homebrew one. This still satisfies the minimum
+required version, so the build will succeed either way; to prefer the Homebrew
+copy instead, pass `-DCMAKE_PREFIX_PATH=$(brew --prefix libxml2)` at configure
+time.
+
+
```
dbear496: This entire section is redundant. I assume users know how to use their own computer, so Mac user's don't need me to tell them how to use brew on their Mac just like Debian users don't need me to tell them to use apt, and Fedora users don't need me to tell them how to use dnf. In fact, it is not even relevant to me *how* users get these dependencies as long as they are available for LinCity-NG to use.

The only thing not redundant here is the pkg-config dependency, which should be listed with the other build dependencies.

lib/cfgpath/CMakeLists.txt
dbear496: This is comment wordy, and it can be removed entirely.

src/lincity/MapPoint.hpp
dbear496: This comment is wordy and doesn't make any sense. `// for std::hash` will suffice.

CHANGELOG.d/macos-build-fix.md
```diff
@@ -0,0 +1,8 @@
+## Unreleased
+
+###### Internal
+- Fixed the build on macOS with the system Clang/libc++ toolchain.
+
+###### Documentation / Translation
+- Documented macOS build dependencies, including a Homebrew naming pitfall
+  for libxml++.
```
dbear496: Using a version formula in homebrew is not a "naming pitfall".

doc/DEPENDENCIES.md
```diff
@@ -79,6 +79,29 @@ Following is the list of dependencies for building and/or running LinCity-NG.
...
+brew install cmake pkg-config sdl3 sdl3_image sdl3_mixer sdl3_ttf \
+  libxml2 libxml++@5 fmt gettext
+```
+
+zlib and xsltproc do not need to be installed separately; both already ship
+with macOS.
+
+Note the `@5`: Homebrew's plain `libxml++` formula is the incompatible 2.x
+series. `brew install libxml++` alone will not satisfy this project's
+`libxml++-5.0` requirement.
```
dbear496: The documentation above already states that libxml++ version 5 is required. Explaining how to use homebrew version formulae is outside the scope of this documentation.

dbear496 (changes requested):
Your AI is quite wordy.

dbear496:
It seems `FSFindFolder` and `FSRefMakePath` are deprecated. The discussion [here](https://github.com/sago007/PlatformFolders/pull/16) and [here](https://github.com/sago007/PlatformFolders/issues/14) may be relevant. Potentially an upstream patch to cfgpath is in order.

src/lincity/MapPoint.hpp
kelvSYC: IWYU says that std::hash is in `<memory>` and not `<string_view>`, but even then, the C++ standard calls for `<functional>`. We can change it to `<memory>` if necessary, but `<string_view>` won't work on macOS at the very least.

doc/BUILD_OPTIONS.md
kelvSYC: Removed.

doc/DEPENDENCIES.md
kelvSYC: Removed.

CHANGELOG.d/macos-build-fix.md
kelvSYC: Removed.

lib/cfgpath/CMakeLists.txt
kelvSYC: Removed.

src/lincity/MapPoint.hpp
kelvSYC: Fixed.

doc/DEPENDENCIES.md
kelvSYC: Removed.

kelvSYC:
> It seems `FSFindFolder` and `FSRefMakePath` are deprecated. The discussion [here](https://github.com/sago007/PlatformFolders/pull/16) and [here](https://github.com/sago007/PlatformFolders/issues/14) may be relevant. Potentially an upstream patch to cfgpath is in order.

We've replaced that code to avoid the use of the Carbon APIs, making it so that CoreServices is no longer needed.

kelvSYC (commented):
Addressed each of the comments.

src/lincity/MapPoint.hpp
dbear496: Could you point me to the place where the standard calls for `<functional>`? IWYU definitely prefers `<string_view>` on my end, so it probably needs to be overridden in mk/iwyu.map.

lib/cfgpath/cfgpath.h
dbear496, line 1: This file is from [cfgpath](https://github.com/Malvineous/cfgpath), so these changes will be overwritten if/when it is updated. Preferably, changes to this file should be made upstream.

lib/cfgpath/cfgpath.h
```diff
@@ -265,10 +278,21 @@ static inline void get_user_config_folder(char *out, unsigned int maxlen, const
 	mkdir(out);
 	strcat(out, "\\");
 #elif defined(CFGPATH_MAC)
-	FSRef ref;
-	FSFindFolder(kUserDomain, kApplicationSupportFolderType, kCreateFolder, &ref);
+	char *home_env = getenv("HOME");
```
dbear496: Name this `home`, and copy both it and `app_support` directly into `out`. There is no good reason to copy in two stages--you need to review the output of your AI before giving it to me.

lib/cfgpath/cfgpath.h
```diff
@@ -265,10 +278,21 @@ static inline void get_user_config_folder(char *out, unsigned int maxlen, const
...
+	if (!home_env) {
+		// Can't find home directory
+		out[0] = 0;
+		return;
+	}
+	const char *app_support = "/Library/Application Support";
```
dbear496: You use raw slashes here but `PATH_SEPARATOR_STRING` later. You should choose one or the other.

lib/cfgpath/cfgpath.h
dbear496, line 279: This appends a constant string to a constant string.

lib/cfgpath/cfgpath.h
```diff
@@ -59,7 +59,9 @@
 #define PATH_SEPARATOR_STRING "\\"
 #elif defined(__APPLE__)
 #define CFGPATH_MAC
-#include <CoreServices/CoreServices.h>
+#include <limits.h>
```
dbear496: Is this for `PATH_MAX`?

dbear496 (changes requested):

src/lincity/MapPoint.hpp
```diff
@@ -27,9 +27,7 @@
 #include <cstddef>        // for size_t
 #include <ostream>        // for ostream
 
-namespace std {
-template<class Key> struct hash; // IWYU pragma: keep
-}  // namespace std
+#include <functional>  // for std::hash
```
dbear496: This should be grouped with the other system includes.

lib/cfgpath/cfgpath.h
dbear496, line 170: This puts the config file directly in ~/Library/Application Support/, which is a no-no according to the Apple documenation [here](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html) and [here](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html).

src/lincity/MapPoint.hpp
kelvSYC: The standard dictates std::hash to be in `<functional>`, according to http://eel.is/c++draft/functional.syn.

This is definitely a portability issue, as the existing code is said to work when using libstdc++ (which is used on Linux), but fails with libc++ (which is used on macOS) due to the way the standard library is implemented causing an ambiguous declaration.

Regardless, we will commission a change to iwyu.map to make this behavior consistent.

lib/cfgpath/cfgpath.h
kelvSYC: According to some research, https://github.com/Malvineous/cfgpath/pull/16 would address the issue, but has been stuck in purgatory. The maintainers there appear to reject the proper practices outlined in Apple's documentation for the sake of backwards compatibility.

Given the desire to treat cfgpath.h as effectively immutable, I'll commission a way to work around the issue rather than through it.

lib/cfgpath/cfgpath.h
kelvSYC: This next set of changes will leave cfgpath.h untouched, as requested.

lib/cfgpath/cfgpath.h
kelvSYC: This next set of changes will leave cfgpath.h untouched, as requested.

lib/cfgpath/cfgpath.h
kelvSYC: This next set of changes will leave cfgpath.h untouched, as requested.

src/lincity/MapPoint.hpp
kelvSYC: This next set of changes will leave cfgpath.h untouched, as requested.

src/lincity-ng/Config.cpp
dbear496, line 1: Please no. Both your prior proposals were better than this. All I said was "It seems `FSFindFolder` and FSRefMakePath` are deprecated. ... Potentially an upstream patch to cfgpath is in order." And now you've recreated cfgpath inside Config.cpp.

Just link the needed framework in lib/cfgpath/CMakeLists.txt, and add a **brief** comment that this is only for deprecated symbols. It would be great to get a patch upstream. I don't think Malvineous/cfgpath#16 is stuck--the contributor just didn't respond to the maintainer's comments. The maintainer didn't even have an objection to removing the deprecated uses.

W.r.t. the config file location, this is also an upstream issue that I believe should be handled upstream rather than here. It looks like the cfgpath project just needs a good way to handle backward-incompatible changes.

doc/DEPENDENCIES.md
```diff
@@ -61,6 +61,12 @@ Following is the list of dependencies for building and/or running LinCity-NG.
 
   xsltproc is provided by LibXslt
 
+* pkg-config
+
+  https://gitlab.freedesktop.org/pkg-config/pkg-config
```
dbear496: I prefer linking to the homepage (when it exists) instead of the source repo: https://www.freedesktop.org/wiki/Software/pkg-config

dbear496 (changes requested):

src/lincity-ng/Config.cpp
kelvSYC: Noted. Rolling it back to how it was in the beginning.

Perhaps asking for the intent of the comment was in order in retrospect. For a first-time contributor, the phrasing of the feedback did not make a distinction between an advisory note or a real merge-blocking change, and I personally read the feedback as "suggestions are not optional". This reading meant treating "avoid the deprecated calls" as absolute, leading to the conclusion that cfgpath is wholly unusable on macOS and that a workaround was necessary.

doc/DEPENDENCIES.md
kelvSYC: Done.

src/lincity/MapPoint.hpp
dbear496: The standard dictates that `std::hash` be declared in a slew of other headers too, e.g. [`<memory>`](https://eel.is/c++draft/memory) and [`<string_view>`](https://eel.is/c++draft/string.view). According to [cppreference](https://en.cppreference.com/cpp/utility/hash), there are 15 different standard headers that declare `std::hash`.

I agree forward-declaring `std::hash` in MapPoint.hpp is not portable; in fact, forward-declaring anything from the standard library is typically ill-advised. And I suppose using `<functional>` here is as good as any.

dbear496 (approved):

src/lincity-ng/Config.cpp
dbear496: > For a first-time contributor, the phrasing of the feedback did not make a distinction between an advisory note or a real merge-blocking change

Sorry about that. I could have been more clear.

PR #393, opened 2026-09-18
remove unused or unread variables

FabioPedretti:


src/lincity-ng/GameView.hpp
```diff
@@ -116,7 +116,6 @@ class GameView : public Component
     void preReadImages(void);
     Texture* readTexture(const std::filesystem::path& filename);
     //void loadTextures();
-    //void preReadCityTexture(int textureType, const std::string& filename);
```
dbear496, line 119: Why not both lines?

dbear496 (dismissed):

src/lincity-ng/GameView.hpp
FabioPedretti: I removed just the second line as it referenced the textureType variable being removed in GameView.cpp.
Since loadTextures is nowhere referenced I removed also this line.
There is other commented code which may be eventually removed.

dbear496 (approved):

